PR #5, opened 2018-05-06
Posenet - changes for feedback and new API

oveddan:
* Pointing to tensorflow/tfjs 
* Fixed some formatting
* Added missing licenses
* Validating output stride
* Fixed global-transform issue
* Validating output stride
* Switched checkpoints to point to the new url.
* Changed architectures to use the multiplier concept. Now they are 1.01 1.00 0.75 0.50
* Change loading posenet to happen in one line via a `load()` call.  This load call now takes a multiplier argument which affects which checkpoint is loaded
* Refactored posenet to take an image as an input, and handle the scaling and reversing.  It scales the outputs back to match the input image resolution.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/5)


posenet/README.md
```diff
@@ -57,13 +57,16 @@ All keypoints are indexed by part id.  The parts and their ids are:
...
+const pose = await poseNet.estimateSinglePose(image, inputSize, reverse, outputStride);
 ```
 
 #### Inputs
 
-* **image** - a 3d tensor of the image data to predict the poses for.
-* **outputStride** - the desired stride for the outputs when feeding the image through the model.  Must be 32, 16, 8.   The higher the number, the faster the performance but slower the accuracy, and visa versa.
+* **image** - ImageData|HTMLImageElement|HTMLCanvasElement|HTMLVideoElement
+   The input image to feed through the network.
+* **inputSize** - The size to scale the image to before feeding through the network. Defaults to 513.  Must have a value which when 1 is subtracted from it, is divisible by the output stride. Sample acceptable values are 29, 161, 193, 257, 289, 321, 353, 385, 417, 449, 481, 513. Set this number lower to scale down the image and increase the speed when feeding through the network.
```
oveddan: I'm wondering if we should change this to a scale factor, and then we can automatically adjust the scale factor to be one of these values.

nsthorat:
Review status: 0 of 15 files reviewed at latest revision, 1 unresolved discussion.

---

*[posenet/package.json, line 13 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/5#-LBs0NVy6vNDc4Cy19Qn:-LBs0NVz2tuEmnBvrKJB:b-i3jmb3) ([raw file](https://github.com/tensorflow/tfjs-models/blob/afcbb2331da5a4794dde150f2deccef1b1c6ab86/posenet/package.json#L13)):*
> ```JSON
>   },
>   "peerDependencies": {
>     "@tensorflow/tfjs": "0.8.1"
> ```

0.10.3 here and below

---

*[posenet/README.md, line 67 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/5#-LBrOuVvB97v0_6n689U-r1-67:-LBs0pgg0dMojBehCExx:bm4cehp) ([raw file](https://github.com/tensorflow/tfjs-models/blob/44d2f8cfa5cfae4113db75bbe4ecc5105bdc0242/posenet/README.md#L67)):*

I'm wondering if we should change this to a scale factor, and then we can automatically adjust the scale factor to be one of these values.
</blockquote></details>

I think that makes sense!

---

*[posenet/README.md, line 276 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/5#-LBs3eAIEGhP-bqbxSDW:-LBs3eAIEGhP-bqbxSDX:b-zd4cml) ([raw file](https://github.com/tensorflow/tfjs-models/blob/afcbb2331da5a4794dde150f2deccef1b1c6ab86/posenet/README.md#L276)):*
> ```Markdown
> 
> ```javascript
> import * as posenet from '@tensorflow-models/posenet';
> ```

can you add a block for non-es6 users? script src users, those users will need to also import tfjs with a script tag.

---

*[posenet/src/mobilenet.ts, line 18 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/5#-LBs3t-QDLye4sqYWQHP:-LBs3t-QDLye4sqYWQHQ:b-sipdhl) ([raw file](https://github.com/tensorflow/tfjs-models/blob/afcbb2331da5a4794dde150f2deccef1b1c6ab86/posenet/src/mobilenet.ts#L18)):*
> ```XML
>  */
> 
> ```

use tfjs here

---

*[posenet/src/posenet.ts, line 37 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/5#-LBs2TvKBiE_AjjXY0c3:-LBs2TvKBiE_AjjXY0c4:bhmzei8) ([raw file](https://github.com/tensorflow/tfjs-models/blob/afcbb2331da5a4794dde150f2deccef1b1c6ab86/posenet/src/posenet.ts#L37)):*
> ```XML
>   const imageTensor = tf.fromPixels(input);
> 
>   if (reverse)
> ```

use {} here

---

*[posenet/src/posenet.ts, line 218 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/5#-LBs2PMW3dKduC0Qvh6z:-LBs2PMW3dKduC0Qvh7-:b-qbf8ac) ([raw file](https://github.com/tensorflow/tfjs-models/blob/afcbb2331da5a4794dde150f2deccef1b1c6ab86/posenet/src/posenet.ts#L218)):*
> ```XML
>           const inputTensor = toInputTensor(input, resolution, reverse);
>           return this.predictForMultiPose(inputTensor, outputStride);
>         })
> ```

semicolon

---

*[posenet/src/posenet.ts, line 243 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/5#-LBs1q-CBjbYF13psoyy:-LBs1q-CBjbYF13psoyz:b9dfbqq) ([raw file](https://github.com/tensorflow/tfjs-models/blob/afcbb2331da5a4794dde150f2deccef1b1c6ab86/posenet/src/posenet.ts#L243)):*
> ```XML
>  * architecture specified by the multiplier.
>  *
>  * @param multiplier An optional string with values: "1.01", "1.00, ""0.75", or
> ```

update this to not a string

---

*[posenet/src/posenet.ts, line 251 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/5#-LBs2vhm4yUgOE6YbHb2:-LBs2vhm4yUgOE6YbHb3:b-pi9d9n) ([raw file](https://github.com/tensorflow/tfjs-models/blob/afcbb2331da5a4794dde150f2deccef1b1c6ab86/posenet/src/posenet.ts#L251)):*
> ```XML
>  * of accuracy.
>  *
>  * @return
> ```

remove this

---

*[posenet/src/posenet.ts, line 253 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/5#-LBs1nOI6aRsdtwMdLzK:-LBs1nOI6aRsdtwMdLzL:b-yhiz7t) ([raw file](https://github.com/tensorflow/tfjs-models/blob/afcbb2331da5a4794dde150f2deccef1b1c6ab86/posenet/src/posenet.ts#L253)):*
> ```XML
>  * @return
>  */
> export default async function load(multiplier: MobileNetMultiplier = 1.01):
> ```

no need to make this a default export

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/5)*

nsthorat:
:lgtm_strong:

---

Review status: 0 of 15 files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/5#-:-LBs40j1EUEJ3v9mXYhn:b-jbsf8z)*

PR #16, opened 2018-05-09
Update coco.js

hpssjellis:
One code change allows the loading and analysis of an image as soon as the image is chosen from the dropdown list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/16)

### DIFF
--- a/posenet/demos/coco.js
+++ b/posenet/demos/coco.js
@@ -221,11 +221,9 @@ function setupGui(net) {
     // in the neural network. The lower the value of the output stride the higher the accuracy 
     // but slower the speed, the higher the value the faster the speed but lower the accuracy.
     gui.add(guiState, 'outputStride', [32, 16, 8])
-        .onChange((outputStride) => guiState.outputStride =
-            Number(outputStride));
+        .onChange(() => testImageAndEstimatePoses(net));
     gui.add(guiState, 'image', images)
         .onChange(() => testImageAndEstimatePoses(net));
-    gui.add(guiState, 'detectPoseButton');
 
     // Pose confidence: the overall confidence in the estimation of a person's 
     // pose (i.e. a person detected in a frame)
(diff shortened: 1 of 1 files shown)


oveddan:
Thanks for this - I agree and it will help with the UI. 

Could you also:
- [ ] Make it so that when the output stride dropdown is changed, the pose estimation also happens
- [ ] Get rid of the estimate pose button right below.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/16#-:-LC79CghCMwdtRIrqVBg:b1wamfg)*

hpssjellis:
Made both extra changes.

Output Stride now works but the lower value of 8 seems to get poor results and the higher value of 32 seems to get better results without much detectable time difference.

Does not seem to match the notes on the page that say

_"The output stride and input image resolution have the largest effects on accuracy/speed. A higher output stride results in lower accuracy but higher speed. ..."_

oveddan:
Thanks for the change.

The issue is that when the output stride dropdown is changed, the output stride becomes a string.  If you see the previous code we converted this to a number, so you should do the same.

Also - can you please do a rebase from master instead of a merge?

`git rebase origin/master`

instead of `git merge master`

It is better to just have the commit with your changes in this PR than also having the merge commit.

You can fix this by squashing the commits.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/16#-:-LCBCSWQ1OHHeYedvtOw:b-78ok8w)*

hpssjellis:
I only use the browser version of git, so easier for me to close this PR and make a new one.

PR #21, opened 2018-05-21
Alert user when getUserMedia throws an error

iloahz:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/21)

### DIFF
--- a/posenet/demos/camera.html
+++ b/posenet/demos/camera.html
@@ -28,6 +28,8 @@
 </head>
 
 <body>
+    <div id="info" style='display:none'>
+    </div>
     <div id="loading">
         Loading the model...
     </div>
--- a/posenet/demos/camera.js
+++ b/posenet/demos/camera.js
@@ -288,10 +288,10 @@ export async function bindPage() {
   try {
     video = await loadVideo();
   } catch(e) {
-    const errorMessage = "This browser does not support video capture, or this device does not have a camera";
-    alert(errorMessage);
-    console.error(e);
-    return;
+    let info = document.getElementById('info');
+    info.textContent = "this browser does not support video capture, or this device does not have a camera";
+    info.style.display = 'block';
+    throw e;
   }
 
   setupGui([], net);
--- a/posenet/demos/camera.js
+++ b/posenet/demos/camera.js
@@ -40,31 +40,29 @@ function isMobile() {
  *
  */
 async function setupCamera() {
+  if (!navigator.mediaDevices || !navigator.mediaDevices.getUserMedia) {
+    throw 'Browser API navigator.mediaDevices.getUserMedia not available';
+  }
+
   const video = document.getElementById('video');
   video.width = videoWidth;
   video.height = videoHeight;
 
-  if (navigator.mediaDevices && navigator.mediaDevices.getUserMedia) {
-    const mobile = isMobile();
-    const stream = await navigator.mediaDevices.getUserMedia({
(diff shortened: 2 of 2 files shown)


nsthorat:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[posenet/demos/camera.js, line 291 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/21#-LD1wE_QDiCF6smUqA1m:-LD1wE_QDiCF6smUqA1n:b-6wluqm) ([raw file](https://github.com/tensorflow/tfjs-models/blob/ad066e6eea73987e204b2db895fb9a00f4dfad3a/posenet/demos/camera.js#L291)):*
> ```JavaScript
>     video = await loadVideo();
>   } catch(e) {
>     const errorMessage = "This browser does not support video capture, or this device does not have a camera";
> ```

Instead of alerting here, would you put an element in the DOM that has this message? Typically we try to stay away from "alert" :)

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/21)*

oveddan:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[posenet/demos/camera.js, line 291 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/21#-LD1wE_QDiCF6smUqA1m:-LD4rzBQBZnQ14h3X4k8:b-2wzwns) ([raw file](https://github.com/tensorflow/tfjs-models/blob/ad066e6eea73987e204b2db895fb9a00f4dfad3a/posenet/demos/camera.js#L291)):*

Instead of alerting here, would you put an element in the DOM that has this message? Typically we try to stay away from "alert" :)
</blockquote></details>

We actually already were alerting.  But I agree changing this to an error element would be preferable.

Looking at the existing code it is odd how it returns a rejected promise, but then the error is caught and the logged with `console.error` instead of just  bubbling up/being re-thrown.

So I think the right thing would be to show an error div, and then just re-throw the original error.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/21)*

iloahz:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[posenet/demos/camera.js, line 291 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/21#-LD1wE_QDiCF6smUqA1m:-LDAPcAn5Odsu-u6VaTj:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/ad066e6eea73987e204b2db895fb9a00f4dfad3a/posenet/demos/camera.js#L291)):*

We actually already were alerting.  But I agree changing this to an error element would be preferable.

Looking at the existing code it is odd how it returns a rejected promise, but then the error is caught and the logged with `console.error` instead of just  bubbling up/being re-thrown.

So I think the right thing would be to show an error div, and then just re-throw the original error.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/21)*

nsthorat:
:lgtm_strong:

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/21#-:-LDEDvEV7yiGIVBHSWUf:b-jbsf8z)*

PR #61, opened 2018-08-17
speech-commands: Add logic for suppression; Improve demo and README

caisq:
- Add logic for suppression and optional parameter `suppressionTimeMillis`
- Add logic for ignoring `_background_noise_` and `_unknown_` tokens
- Add optional parameter `invokeCallbackOnNoiseAndUnknown`
- Switch to model that includes unknown tokens
- Update README
- Respond to @dsmilkov 's comments on an earlier PR
- demo page: Add highlighting of the word with top score

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/61)


caisq:
Thanks for the reviews, @dsmilkov and @pyu10055

PR #63, opened 2018-08-19
Person and Body Part Segmentation model

oveddan:
This PR adds support for loading and decoding new pre-trained models provided by Google researcher Tyler Zhu (@tylerzhu-github) which does person segmentation and body part mapping on an image.  It also adds demos showing how they can be visualized.

The new demos can be tried here:
https://person-segmentation-demos.netlify.com

Here is segmentation and estimated part map with output stride of 8:  The dark gray area indicates what is outside of the segmentation area:
download 3
download 4

Here is purely a segmentation output with output stride of 16.  The dark gray area indicates what is outside of the segmentation area:
download 2

## The Model

This new model provides two outputs: a 2d tensor for segmentation, and a 3d tensor used for body part maps.  It requires an image size of 257x353.   It only supports segmentation on a single person. However, if combined with a person box detector in a first stage, it can be used in a second stage to do segmentation and part maps for each bounding box, therefore becoming a multi-person segment and part mapper estimator.  

Unlike Posenet's pose estimation, which has 17 keypoints, the part map refers to 24 *part channels*:

```typescript
export const partChannels: string[] = [
  'left_face',
  'right_face',
  'right_upper_leg_front',
  'right_lower_leg_back',
  'right_upper_leg_back',
  'left_lower_leg_front',
  'left_upper_leg_front',
  'left_upper_leg_back',
  'left_lower_leg_back',
  'right_feet',
  'right_lower_leg_front',
  'left_feet',
  'torso_front',
  'torso_back',
  'right_upper_arm_front',
  'right_upper_arm_back',
  'right_lower_arm_back',
  'left_lower_arm_front',
  'left_upper_arm_front',
  'left_upper_arm_back',
  'left_lower_arm_back',
  'right_hand',
  'right_lower_arm_front',
  'left_hand'
];
```

The 3d tensor output for body part map's 3rd axis corresponds to each of these part channels, and for each part channel has a 2d tensor which indicates the probability that a part channel exists in each area.

There are 4 MobileNet-architecture models available as part of this PR, with 1.00, 0.75, 0.50, and 0.25 depth multipliers.  The higher the depth multiplier the higher the accuracy and slower the speed, and visa versa with lower depth multipliers.

These new models are in the frozen graph format, so the PoseNet code was modified to support loading weights either from a frozen graph format or the legacy checkpoint format which is what the existing posenet models use.

The model is loaded like:

```typescript
export async function load(multiplier: MobileNetMultiplier = 0.75)
```

Unlike the existing PoseNet `load` function, this loads using the frozen graph format. 

## The new API

This new model, called PersonSegmentation, provides three api methods:

```typescript
async estimatePersonSegmentation(
      input: InputType, flipHorizontal = false, outputStride: OutputStride = 16,
      segmentationThreshold = 0.5): Promise<Int32Array>

async estimatePartSegmentation(
      input: InputType, flipHorizontal = false, outputStride: OutputStride = 16,
      segmentationThreshold = 0.5,
      partColors: Array<[number, number, number]>): Promise<Int32Array>
```

The segmentation mask returned from `estimatePersonSegmentation` is an array with values 0 or 1 corresponding to if a person was estimated to be in each pixel. The array size corresponds to the number of pixels in the image. See the readme for the body part ids.  If a value is 1, a person was estimated to be in the corresponding pixel.

The returned value from estimatePartSegmentation is an array with values 0-24 corresponding to the body part for each pixel, indexed by pixel h, w.  The array size corresponds to the number of pixels in the image. See the readme for the body part ids.  If a value is -1, no body part was estimated to be in that pixel.

The model handles resizing and padding the image to match the **required resolution of 257x353** while maintaining the aspect ratio before feeding it through the network.  It also scales the outputs back up to the original input resolution while cropping out the padding before returning the outputs.  

## The demos

This includes new demos for segmentation and part maps that only work with a single person.

The demos can be viewed here:
https://person-segmentation-demos.netlify.com

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/63)


posenet/src/posenet_model.ts
```diff
@@ -167,13 +204,97 @@ export class PoseNet {
...
+    const [height, width] = getInputTensorDimensions(input);
+    const resizedHeight =
+        getValidResolution(imageScaleFactor, height, outputStride);
+    const resizedWidth =
+        getValidResolution(imageScaleFactor, width, outputStride);
+
+    const {heatmapScores, offsets, segmentScores, partHeatmapScores} =
+        tf.tidy(() => {
+          const inputTensor =
+              toInputTensor(input, resizedHeight, resizedWidth, flipHorizontal);
```
tyleeer: The new model works best with input crop of size 353x257 (pixels), as that's what it was trained on, so it is necessary to add a functionality to crop a sub region of the `inputTensor` and resize the sub-region to 353x257 pixels before feeding into the network. To prevent introducing aspect-ratio distortion, the sub-region should have the aspect ratio of (353 : 257) before being resized to 353x257 (pixels).

oveddan: Will do.

tylerzhu-github:
Thanks Dan for the PR and it looks good to me. Ideally we could also add the functionality to show all the predictions (part mask, 2D keypoints, instance mask) together for demo usage on a high-end workstation. Other than that this looks great!

nsthorat:
Send me a ping when you are ready for another review!

oveddan:
___
*[posenet/src/index.ts, line 33 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/63#-LOShLxt0oi7-RpP2RWe:-LP1sOQs9d35gCphpbrO:bhwyykn) ([raw file](https://github.com/tensorflow/tfjs-models/blob/b9eb847adbad45588ee7ed16e0ba52297c1b9678/posenet/src/index.ts#L33)):*
<details><summary><i>Previously, dsmilkov (Daniel Smilkov) wrote…</i></summary><blockquote>

Is the newer method faster/more precise in detection poses? If not, let's remove the pose-detection capability from the newer model, so the 2 models are complementary and have no overlaps. Simplifies the story/API/demos.
</blockquote></details>

I'll defer the accuracy question to @tylerzhu-github .  Regarding speed, it's likely to be slower than the existing PoseNet model as the new model requires the image to be a specific size, whereas the existing model allowed the image to be scaled to smaller sizes resulting in a faster speed.  One thing that's nice about the new model is it can do all 3 of pose estimation, segmentation, and part map, so we would be losing a bit of functionality if we drop it from the new model but I can see the big advantage of it being simpler and clearer documentation wise.

Also - if we decide to drop pose estimation this can be a separate model, living in its own folder; all that would be shared between the two models is being able to have a mobilenet model with variable output stride.

oveddan:
___
*[posenet/src/index.ts, line 33 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/63#-LOShLxt0oi7-RpP2RWe:-LP2W9fB6mAP3w38Wvxl:b-quby8o) ([raw file](https://github.com/tensorflow/tfjs-models/blob/b9eb847adbad45588ee7ed16e0ba52297c1b9678/posenet/src/index.ts#L33)):*
<details><summary><i>Previously, dsmilkov (Daniel Smilkov) wrote…</i></summary><blockquote>

I just chatted with Nikhil. Let's drop pose estimation and move into separate model. It will also make it much more discoverable.  Thanks!
</blockquote></details>

ok sounds good - what should it be named? body_segmentation?

tylerzhu-github:
Having the model in its own folder sounds great to me! Thanks for the detailed reviews!

oveddan:
New computer - I committed with the wrong email address before.

tylerzhu-github:
+1 for some unit tests. In addition (not needed in this PR, perhaps add a TODO for me is that: we should also have some regression test (if it's not too slow to run). For example we can include several test images and make sure the predicted segmentation masks value are within some error range. The initial groundtruth segmentation for these images could be generated using the model at the current PR snapshot.

person-segmentation/src/util.ts
```diff
@@ -16,7 +16,6 @@ export function toInputTensor(input: PersonSegmentationInput) {
 export function cropAndResizeTo(
```
tylerzhu-github: Is there a particular reason not to use the Tensorflow.js implementation of crop_and_resize op? (I am thinking their native implementation will be faster?)

https://js.tensorflow.org/api/0.13.0/#image.cropAndResize

tylerzhu-github:
Thank you so much Dan and it is very fast indeed. I think we are good, maybe we can explore GPU rendering later in another PR after this PR is merged. You get lgtm from me!

One thing i notice is the mask and image are not synced (you can see it when subject is moving fast) I did some investigation locally and found it is caused by the model inference and rendering is using different image tensor (mask inference is 1 frame behind) I did some changes and it's synced but at the cost of slowing down the fps (since I basically reused most of the tf.toPixels code to convert tensor to ImageData).

This is the recording of mask and image synced.
http://g.recordit.co/fUSTQN50OQ.gif

Here is the files i modified to achieve this. I am looking for ways to speed it up ( right now i am just the converting the image tensor used by model inference to ImageData and use that for rendering).
camera
output_rendering_util
person_segmentation_model

tylerzhu-github:
Thank you so much Dan and it is very fast indeed. I think we are good, maybe we can explore GPU rendering later in another PR after this PR is merged. You get lgtm from me!

One thing i notice is the mask and image are not synced (you can see it when subject is moving fast) I did some investigation locally and found it is caused by the model inference and rendering is using different image tensor (mask inference is 1 frame behind) I did some changes and it's synced but at the cost of slowing down the fps (since I basically reused most of the tf.toPixels code to convert tensor to ImageData).

This is the recording of mask and image synced.
http://g.recordit.co/fUSTQN50OQ.gif

Here is the files i modified to achieve this. I am looking for ways to speed it up ( right now i am just the converting the image tensor used by model inference to ImageData and use that for rendering).
camera
output_rendering_util
person_segmentation_model

dsmilkov:
See also the travis failure: https://travis-ci.org/tensorflow/tfjs-models/jobs/445232499#L624

oveddan:
___
*[person-segmentation/src/output_rendering_util.ts, line 28 at r12](https://reviewable.io/reviews/tensorflow/tfjs-models/63#-LPMVbJ_-vWXuYOKyZDP:-LPXnsJh8EN8BvmWAiW7:b-z5779s) ([raw file](https://github.com/tensorflow/tfjs-models/blob/f1ece3f3bf4165819830c0d6aa8ca623f93e1220/person-segmentation/src/output_rendering_util.ts#L28)):*
<details><summary><i>Previously, dsmilkov (Daniel Smilkov) wrote…</i></summary><blockquote>

Side note: Its worth trying to use css blurring on the canvas itself. Safari didn't use to support it, but I think that may have changed: canvas.style['-webkit-filter'] = 'blur(2px)'; Would be great to double-check
</blockquote></details>

I tried this but sadly it doesn't work.

This demo does:
https://codepen.io/zhaojun/pen/zZmRQe 

Will try to reproduce

tylerzhu-github:
Thanks Dan,  this looks good to me! Thank you!

It occurs to me that to get better person segmentation bokeh effect, we can use a trick to generate alpha mask blending (instead of binary mask) basically we need use the person mask to have a softened edge.

PR #65, opened 2018-08-24
add cross-env for knn-classifier demo project

code945:
add cross-env otherwise user may got "NODE_ENV is not recognized as an internal or external command" when run yarn watch on windows.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/65)


code945:
I signed it!

code945:
I signed it!

nsthorat:
Thank you for the fix!

PR #75, opened 2018-09-19
tfjs version bump for posenet - massive speed increase for inference

oveddan:
Version bump posenet and its demo to tfjs 0.13.  Thanks to optimizations in tfjs-core, this now gives a 2-3x increase in speed for inference.

Prior to this update I would get 4-5 fps with the 1.00 model running the camera demo on my Macbook Retina.  With this update I get 12-15 fps.

@nsthorat can you link to the commit or PR in tfjs-core that had this optimization in it?

### DIFF
--- a/posenet/demos/package.json
+++ b/posenet/demos/package.json
@@ -9,8 +9,8 @@
     "node": ">=8.9.0"
   },
   "dependencies": {
-    "@tensorflow-models/posenet": "0.1.2",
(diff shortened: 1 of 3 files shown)


nsthorat:
Two PRs that made this faster:
https://github.com/tensorflow/tfjs-core/pull/1249
https://github.com/tensorflow/tfjs-core/pull/1274

PR #85, opened 2018-10-11
[speech-commands] Add support for custom model and metadata URLs

caisq:
With unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/85)


caisq:
cc @pyu10055

PR #99, opened 2018-11-08
[speech-command] Add includeEmbeddings; Allow recognize() to draw from WebAudio directly

caisq:
- Allow `recognize()` to take no argument, in which case the method will draw a frame of
  audio directly from WebAudio.
- Add the config field `includeEmbedding` to `startStreaming()` and `recognize()` methods.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/99)


nsthorat:
Just an FYI regarding default parameters (no need for changes), if you pass `undefined` explicitly it will default to the default you provide in the args. `null` explicitly will not.

PR #100, opened 2018-11-08
[speech-commands] Changes to API and model location

caisq:
- startStreaming() --> listen()
- stopStreaming() --> stopListening()
- isStreaming() --> isListening()
- Update model location

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/100)


nsthorat:
Yes, but it has no weights

PR #106, opened 2018-11-15
[speech-commands] add fine-tuning and demo

caisq:
- Fine-tuning happens after the initial transfer training and unfreezes a Dense layer in the base model
- Demo: require a minimum number of examples per category
- Demo: Allow fine-tuning and plot loss and accuracy values during fine-tuning
- Add util for balanced train-validation splits

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/106)


speech-commands/demo/index.js
```diff
@@ -172,11 +176,28 @@ enterLearnWordsButton.addEventListener('click', () => {
           exampleCanvas, spectrogram.data, spectrogram.frameSize,
           spectrogram.frameSize);
       const exampleCounts = transferRecognizer.countExamples();
+      const classNames = Object.keys(exampleCounts);
+      let minCountByClass;
```
pyu10055: Instead of using classNames length to determine missing, probably better
to use transferWords to iterate through the exampleCounts, and do a reduce on the array.

speech-commands/src/browser_fft_recognizer.ts
```diff
@@ -754,30 +767,70 @@ class TransferBrowserFftSpeechCommandRecognizer extends
...
         callbacks: config.callback == null ? null : [config.callback]
       });
-      tf.dispose([xs, ys]);
-      return history;
-    } catch (err) {
-      tf.dispose([xs, ys]);
-      this.model = null;
-      return null;
+
+      if (config.fineTuningEpochs != null && config.fineTuningEpochs > 0) {
```
pyu10055: is fineTuning phase an example for the chapter or it provides better result if it is done separately after the transfer learning?
And should the layer trainable get reset after the fine tuning is done?

PR #112, opened 2018-11-28
Added windows support for all demos

MrRefactoring:
Without `cross-env NODE_ENV` was impossible to run the demos on windows

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/112)


nsthorat:
Thanks for the PR!

PR #113, opened 2018-11-28
[speech-commands] Change dataset serialization format: 2 files -> 1 file

caisq:
- demo is updated accordingly

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/113)


speech-commands/src/dataset.ts
```diff
@@ -361,23 +412,57 @@ export function deserializeExample(
...
- * @param buffers A number of array buffers to concatenate.
- * @returns Result of concatenating `buffers` in order.
+ * Format of the binary ArrayBuffer:
+ *   1. An 8-byte descriptor (see above).
+ *   2. A 4-byte version number as Uint32.
+ *   3. A 4-byte number for the byte length of the JSON manifest.
+ *   4. The encoded JSON manifest
+ *   5. The binary data of the spectrograms, and raw audio (if any).
+ *
+ * @param serialized: Intemediate serialization format of a dataset.
```
rusheywood:
```suggestion
 * @param serialized: Intermediate serialization format of a dataset.
```

speech-commands/src/generic_utlis_test.ts
```diff
@@ -0,0 +1,34 @@
+
```
rusheywood: Filename maybe has a typo? Should it be "generic_utils_test.ts" rather than "generic_utlis_test.ts"

caisq: Fixed. Thanks for catching that!

PR #116, opened 2018-12-03
[speech-commands] Support collecting longer recordings for transfer learning

caisq:
Rationale: 
- Longer recordings allow temporal jitters, which is a form of data augmentation that has been shown to benefit online-recognition accuracy in empirical tests

API level changes:
- `collectExample()` now has a new config parameters: `durationMultiplier`, which if set to a value >1, will let the method collect a spectrogram longer than that required by the underlying TensorFlow.js model.

### DIFF
--- a/speech-commands/src/dataset.ts
+++ b/speech-commands/src/dataset.ts
@@ -324,11 +324,10 @@ export class Dataset {
           }
 
           const snippetLength = spectrogram.data.length / frameSize;
-          const maxIntensityFrame = currentLabel === BACKGROUND_NOISE_TAG ?
+          const focusIndex = currentLabel === BACKGROUND_NOISE_TAG ?
               null :
-              getMaxIntensityFrameIndex(spectrogram);
+              getMaxIntensityFrameIndex(spectrogram).dataSync()[0];
           // TODO(cais): See if we can get rid of dataSync();
-          const focusIndex = maxIntensityFrame.dataSync()[0];
 
           const windows =
               getValidWindows(snippetLength, focusIndex, numFrames, hopFrames);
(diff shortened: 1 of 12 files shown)


speech-commands/src/dataset.ts
```diff
@@ -84,6 +84,42 @@ export interface SerializedExamples {
   data: ArrayBuffer;
 }
 
+export const BACKGROUND_NOISE_TAG = '_background_noise_';
+
+/**
+ * Configuration for getting spectrograms as tensors.
+ */
+export interface GetSepctrogramsAsTensorsConfig {
```
rusheywood:
```suggestion
export interface GetSpectrogramsAsTensorsConfig {
```

speech-commands/src/dataset.ts
```diff
@@ -209,7 +245,8 @@ export class Dataset {
    *     `Dataset`, or
    *   - if the `Dataset` is currently empty.
    */
-  getSpectrogramsAsTensors(label?: string):
+  getSpectrogramsAsTensors(
+      label?: string, config?: GetSepctrogramsAsTensorsConfig):
```
rusheywood:
```suggestion
      label?: string, config?: GetSpectrogramsAsTensorsConfig):
```

speech-commands/src/index.ts
```diff
@@ -69,7 +69,7 @@ export function create(
   }
 }
 
-export {Dataset} from './dataset';
+export {BACKGROUND_NOISE_TAG, Dataset, GetSepctrogramsAsTensorsConfig, getMaxIntensityFrameIndex, spectrogram2IntensityCurve} from './dataset';
```
rusheywood:
```suggestion
export {BACKGROUND_NOISE_TAG, Dataset, GetSpectrogramsAsTensorsConfig, getMaxIntensityFrameIndex, spectrogram2IntensityCurve} from './dataset';
```

PR #117, opened 2018-12-04
[speech-commands] Let highlighting for recognized word fade after a TTL

caisq:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/117)

### DIFF
--- a/speech-commands/demo/ui.js
+++ b/speech-commands/demo/ui.js
@@ -194,10 +194,12 @@ export function plotPredictions(
       if (word === topWord) {
         candidateWordSpans[word].classList.add('candidate-word-active');
         if (timeToLiveMillis != null) {
-          setTimeout(
-              () => candidateWordSpans[word].classList.remove(
-                  'candidate-word-active'),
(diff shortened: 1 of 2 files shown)


speech-commands/demo/ui.js
```diff
@@ -189,6 +193,12 @@ export function plotPredictions(canvas, candidateWords, probabilities, topK) {
     for (const word in candidateWordSpans) {
       if (word === topWord) {
         candidateWordSpans[word].classList.add('candidate-word-active');
+        if (timeToLiveMillis != null) {
+          setTimeout(
+              () => candidateWordSpans[word].classList.remove(
```
rusheywood: optional: i suggest adding a nullcheck guard here, like "if (candidateWordSpans[word]) {..." in case this timeout logic races against the list of words changing. I realize it's unlikely, but GUI races are annoying to debug, and an unlucky future refactor could make this more probable in the future...

caisq: Good point. Done. Thanks.

PR #121, opened 2018-12-19
[speech-commands] Allow collection of long noise examples

caisq:


speech-commands/src/browser_fft_recognizer.ts
```diff
@@ -670,8 +690,7 @@ class TransferBrowserFftSpeechCommandRecognizer extends
       };
       this.audioDataExtractor = new BrowserFftFeatureExtractor({
         sampleRateHz: this.parameters.sampleRateHz,
-        numFramesPerSpectrogram:
-            Math.round(this.nonBatchInputShape[0] * durationMultiplier),
+        numFramesPerSpectrogram,
```
rusheywood: for my education: why does this work? Typescript allows mixes of positional and named constructor assignment? Consider naming the parameter anyway, for robustness...?

caisq: Ah - this is called object literal and is actually a feature of ECMAScript. TypeScript has this syntax because it's a superset of ECMAScript. See: 
https://dev.to/sarah_chima/enhanced-object-literals-in-es6-a9d

This is just a more succinct way of writing:
`{numFramesPerSpectrogram: numFramesPerSpectrogram}`

caisq: Yes, TS does leverage variable name. This is a literal because the variable name actually becomes the field name in the JSON object.

caisq:
Thank you for the review, @rusheywood !

PR #126, opened 2018-12-28
[speech-commands] Add JavaScript training script for browser-FFT audio model (finally)

caisq:
Accompanying changes:
- Add `shuffle` and `normalize` options to `Dataset.getSpectrogramsAsTensors()`
- Add unit test for `normalize()`

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/126)

PR #127, opened 2019-01-04
[speech-command] Check in responses to review comments from #126

caisq:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/127)


caisq:
@bileschi, sorry i forgot to push in #126 after addressing your comments

PR #133, opened 2019-01-13
[body-pix] - mobile improvements for demo.  Allow switching of the camera

oveddan:
Fix body-pix demo issues on mobile.

Demo with the fixes can be seen [here.](https://body-pix-mobile-fix.netlify.com)

* Allowing the user to change the camera.
* Added a "flipHorizontal" and "showFps" option
* On mobile, use the camera label to select the facing mode.
* Refactor demo state - separate gui state from the demo state.

Fixes [#955](https://github.com/tensorflow/tfjs/issues/955)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/133)


oveddan:
Couple issues that should be fixed in a future PR:
* When switching cameras on IOS, occasionally the demo freezes.  Navigating to the home screen then reopening safari causes it to unfreeze.  I tried to debug this for a while with no luck but we can attempt it again in the future.
* I could not find an elegant way to determine if the camera is a rear camera or not, and if it should be flipped horizontally automatically, so for now flipHorizontal is a checkbox in the gui that should be unchecked when switching to the rear facing mode. This can be something that we can work on in a future improvement.

PR #149, opened 2019-02-14
Fix a parcel build warning; Bump version to 0.2.5

caisq:
The parcel issue is described at:
https://github.com/parcel-bundler/parcel/issues/2363

parcel doesn't like `fs.readFileSync`. This PR uses `fs.readFile` to work around that.

BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/149)


caisq:
Thanks for the review, @dsmilkov

PR #162, opened 2019-03-06
Update speech-commands version to depend on tfjs 1.0 and add publish script.

annxingyuan:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/162)


annxingyuan:
I think it makes sense to hold off on bumping the dependency - there are still quite a few failing tests so I don't want to rush it.

PR #168, opened 2019-03-06
Update models to depend on tfjs 1.0

tafsiri:
Also updates demos to depend on un-yet unpublished demos. So this should probably be merged after new model versions are published.

Models updated are: bodypix, cocossd, knn-classifier, posenet.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/168)

### DIFF
--- a/body-pix/package.json
+++ b/body-pix/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow-models/body-pix",
-  "version": "0.1.1",
+  "version": "0.2.0",
   "description": "Pretrained BodyPix model in TensorFlow.js",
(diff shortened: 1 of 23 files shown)


body-pix/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow-models/body-pix",
-  "version": "0.1.1",
+  "version": "0.2.0",
```
dsmilkov: Update all these models to 1.0.0 since they now expect tfjs 1.0 as peer and most users that depend on 0.x of these models, have tfjs 0.x as their dep

PR #174, opened 2019-03-19
Update versions of tfjs and mobilenet in the example code

charliegerard:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/174)

### DIFF
--- a/mobilenet/README.md
+++ b/mobilenet/README.md
@@ -20,7 +20,7 @@ There are two main ways to get this model in your JavaScript project: via script
 
 ```html
 <!-- Load TensorFlow.js. This is required to use MobileNet. -->
-<script src="https://cdnjs.cloudflare.com/ajax/libs/tensorflow/1.0.1/tf.js"> </script>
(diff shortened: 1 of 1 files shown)


charliegerard:
Just signed the CLA

mobilenet/README.md
```diff
@@ -20,9 +20,9 @@ There are two main ways to get this model in your JavaScript project: via script
 
 ```html
 <!-- Load TensorFlow.js. This is required to use MobileNet. -->
-<script src="https://cdn.jsdelivr.net/npm/@tensorflow/tfjs@0.15.3"> </script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/tensorflow/1.0.1/tf.js"> </script>
```
nsthorat: thanks so much for the PR Charlie!

would you mind reverting this to jsdeliver? we've been trying to use jsdelivr across the board since it gives us good statistics (and no need for the tf.js at the end, just replacing 0.15.3 with 1.0.1 will work for jsdeliver).

thank you!

charliegerard: Sure, no problem, sorry! Doing it now!

PR #175, opened 2019-03-20
Use fetch polyfill for node in universal sentence encoder.

annxingyuan:
This PR fixes https://github.com/tensorflow/tfjs/issues/1427

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/175)

### DIFF
--- a/universal-sentence-encoder/package.json
+++ b/universal-sentence-encoder/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow-models/universal-sentence-encoder",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "description": "Universal Sentence Encoder lite in TensorFlow.js",
   "main": "dist/index.js",
   "jsnext:main": "dist/universal-sentence-encoder.esm.js",
@@ -15,13 +15,18 @@
   "peerDependencies": {
     "@tensorflow/tfjs": "^1.0.0"
   },
+  "dependencies": {
+    "node-fetch": "~2.1.2"
+  },
   "devDependencies": {
     "@tensorflow/tfjs": "^1.0.0",
+    "@types/node-fetch": "^2.1.2",
     "@types/jasmine": "~2.5.53",
     "jasmine": "^3.3.1",
     "jasmine-core": "^3.3.0",
     "rimraf": "~2.6.2",
     "rollup": "~0.58.2",
+    "rollup-plugin-commonjs": "9.1.3",
     "rollup-plugin-node-resolve": "~3.3.0",
     "rollup-plugin-typescript2": "~0.13.0",
     "rollup-plugin-uglify": "~3.0.0",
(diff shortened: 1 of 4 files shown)


alexandrduduka:
Hello guys, thanks for your great effort on this library!

I would like to ask if there is a potential of progress on this issue in close future? I see that you did some work to use node-fetch library in tfjs package and updated dependencies recently, however I am currently facing following issues when trying to work with package in node.js environment:
a) fetch is not defined in USE (tfjs 1.1.0, @tensorflow-models/toxicity 1.0.2)
b) typescript is barking during compilation, as it can't find browser apis from dependencies (I guess this can be resolved with ts-ignore on my side.

I am also not sure if it is optimal to use this package on server side, as it is based on tfjs-core, which is not optimised for such use, as far as I understood. Do you think it is better solution to prepare alternative version of library based on tfjs-node, or is it possible to do something with current library on my side?
I am quite new to tensorflow, so if I am not taking anything right, please let me know.

alexandrduduka:
@tafsiri, @annxingyuan, hello, thank you for your great effort on this library!

I would like to ask if there is a potential of progress on this issue in close future? I see that you did some work to use node-fetch library in tfjs package and updated dependencies recently, however I am currently facing following issues when trying to work with package in node.js environment:
a) fetch is not defined in USE (tfjs 1.1.0, @tensorflow-models/toxicity 1.0.2)
b) typescript is barking during compilation, as it can't find browser apis from dependencies (I guess this can be resolved with ts-ignore on my side.

I am also not sure if it is optimal to use this package on server side, as it is based on tfjs-core, which is not optimised for such use, as far as I understood. Do you think it is better solution to prepare alternative version of library based on tfjs-node, or is it possible to do something with current library on my side?
I am quite new to tensorflow, so if I am not taking anything right, please let me know.

annxingyuan:
Hi @alexandrduduka - apologies for the delay on this PR - I will try to move this forward today.

annxingyuan:
@alexandrduduka actually we're about to republish the models with our new fetch utility which solves this problem: https://github.com/tensorflow/tfjs-models/pull/194

alexandrduduka:
Thanks for the fast reply @annxingyuan! I feel like I am using new version which should have it resolved (if I take it correctly PR you mentioned was merged yesterday), but it still doesn't make a trick for me :-( 

Also, could you elaborate a bit on my question about running the package on server side - do you think it is viable solution or do I have to make some fork with tfjs-node used? Maybe there is a way to tell the library to use tfjs-node instead of tfjs-core?

annxingyuan:
ah you're right - that PR updated the dependency to a version of core with the fetch utility but we still need to invoke it - I submitted a follow up PR: https://github.com/tensorflow/tfjs-models/pull/196

you should be able to just run the package server-side (once this new PR is merged)

alexandrduduka:
Thank you! 👍

PR #182, opened 2019-03-30
Make sure to use MobileNetV2

Lewuathe:
MobileNetV2 should be available by passing version parameter.

```ts
src/index_test.ts:41:34 - error TS2345: Argument of type '2' is not assignable to parameter of type '1'.

41     const mobilenet = await load(2);
```

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/182)


jenperson:
This code change has since been made, so I am closing the pull request.

PR #187, opened 2019-04-09
[posenet] Add smoothing pose anomolies option by averaging sequential frames

shopped:
Sometimes an error in joint inference can cause a pose visualization to look janky which looks broken. A common fix I have implemented (and see it implemented in the “move mirror” demo,) is to estimate the pose over time by taking the average of previous positions, providing a smoothing effect that provides steady motion with lag as a tradeoff.

I have added the logic for this in the demo project, with an option to turn it on/off to varying degrees. Hopefully other users of this model will find it helpful. Keep in mind this only works in “single-pose”

Here is a video demo of the changes
https://youtu.be/Olnx16XvWac

I'm open to any feedback maintainers! Love this project

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/187)

### DIFF
--- a/body-pix/demos/.babelrc
+++ b/body-pix/demos/.babelrc
@@ -12,7 +12,7 @@
       }
     ]
   ],
-  "plugins": [
-    "transform-runtime"
-  ]
+  // "plugins": [
+  //   "transform-runtime"
+  // ]
 }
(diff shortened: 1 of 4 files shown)


shopped:
This PR is stale, and I included other local changes. Going to re-open with a cleaner version

PR #188, opened 2019-04-10
Allow users of knn to use arbitrary class Id

dsmilkov:
Fixes https://github.com/tensorflow/tfjs/issues/1478

Users can now pass arbitrary classId/label when calling `addExample()`. The classId/label can be an arbitrary string or number.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/188)

### DIFF
--- a/knn-classifier/README.md
+++ b/knn-classifier/README.md
@@ -7,7 +7,7 @@ algorithm.
 This package is different from the other packages in this repository in that it
 doesn't provide a model with weights, but rather a utility for constructing a
 KNN model using activations from another model or any other tensors you can
-associate with a class.
+associate with a class/label.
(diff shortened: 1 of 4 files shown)


dsmilkov:
Yup, just fixed the bug from before. Strings were never allowed. I can easily allow strings, but it will be strange since classifier.predictClass() will return
```js
{
  classIndex: 'label1',
  confidences: ['label1': 0.9, 'another_label': 0.05,....]
}
```
I wish I could rename `classIndex` to `label` but not worth it. If strings are important to you and will simplify your code, I can relax the constraint but keep `classIndex` in the result signature. Let me know.

PR #189, opened 2019-04-15
Update version of posenet

dsmilkov:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/189)


dsmilkov:
Released new version of posenet, updated the demo to depend on it, tested it and uploaded the new demo on GCS

PR #202, opened 2019-05-01
Remove version from tfjs script tag

xmbrst:
It looks like the version number  in the tfjs script tag url is producing an error because of the fix for issue #196.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/202)


jenperson:
It looks like this change has now been made to the README.

PR #221, opened 2019-05-29
[WIP] PoseNet - load new mobilenet graph model.

oveddan:
Changed posenet mobilenet models to load graph models.

* got rid of all the custom inference code.
* removed checkpoint loader

Still to do:
* Load different models based on output stride and multiplier from different urls, which should be hosted.

To discuss: can we remove the requirement for input image resolution on the mobilenet model since it is a hyperparameter?

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/221)


dsmilkov:
Superseded by #227

PR #226, opened 2019-05-31
[speech-commands] Enable creating recognizer from artifacts

caisq:
Previously, to load a custom model, one has to point `speechCommands.create()` at two URLs: a model.json  URL and a metadata.json URL. This PR changes the two arguments to `speechCommands.create()` so that they can be actual model artifacts and metadata objects instead. This facilitates loading of models from non-URL based routes such as file uploads in browsers.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/226)


caisq:
cc @irealva @hapticdata

hapticdata:
looks good to me! I like this solution, more versatile than merely supporting `tf.io.browserFiles`

PR #229, opened 2019-06-04
Add DeepLab

sdll:
This PR adds the DeepLab model together with a demo.

~~There are several issues with the implementation at the moment.~~ Here is the todo list:

- [x] **Fix the grayscale-only segmentation maps**

~~The culprit might be [here](https://github.com/tensorflow/tfjs-models/blob/5fa0787a968799f0a614c663f83afb09b7e8f2cb/deeplab/src/utils.ts#L86).~~

**UPDATE 05/06**: The problem was resolved by constructing a single buffer for the translated segmentation map, instead of building the tensor by stacking three separate channel buffers (see [here](https://github.com/tensorflow/tfjs-models/commit/218383a59dd84072dfb9e0d2ad6ec83e8b3d45d4)). I do not quite understand why this solves the problem.

- [x] **Show performance stats in the demo**

- [x] **Add loader for improved UX**

- [x] **~~Fix~~ ~~Disable~~ Reserve `tf` to make mangling work**

The code uses ES6 features, so I have picked [`rollup-plugin-terser`](https://github.com/TrySound/rollup-plugin-terser) instead of uglify, which breaks on the compilation of the esm module. ~~Terser, however, is pestered with bugs when mangling is enabled: the demo fails to recognise the minified `tf.tidy` function (the error is, say, `t.tidy is not a function`). I might as well revert to uglify after refactoring the code.~~

**UPDATE 06/06**: ~~The problem persists even after playing with the mangling options, with the model missing `predict` method in the debugging mode. I have disabled mangling altogether and reported the issue (https://github.com/terser-js/terser/issues/364).~~

**UPDATE 29/06**: Mangling works well when `tf` is reserved. See the awesome [explanation](https://github.com/terser-js/terser/issues/364#issuecomment-506958587) by @fabiosantoscode why this might be the case.

- [x] **Improve docs**

- [x] **Add tests**

- [x] **Explore quantisation**

**UPDATE 07/06**: The models are quantized by default.

- [x] Add Cityscapes and ADE20K Labelling Schemes
- [x] Change the gif demo to show all models with updated color mappings

Please let me know if I am missing anything.

cc: @manrajgrover

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/229)


deeplab/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+**Note**: *For more granular control, consider `segment` and `translate` methods described below.*
+
+### Producing a Semantic Segmentation Map
+
+To segment an arbitrary image and generate a two-dimensional tensor with class labels assigned to each cell of the grid overlayed on the image (with the maximum number of cells on the side fixed to 513), use the `segment` method of the `SemanticSegmentation` object.
+
+#### `model.segment` input
+
+* **image** :: DeepLabInput | ImageData | HTMLImageElement | HTMLCanvasElement | HTMLVideoElement | tf.Tensor3D;
```
manrajgrover: Let's code quote types here

deeplab/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+#### `model.segment` output
+
+* **rawSegmentationMap** :: Promise<RawSegmentationMap | tf.Tensor2D>
```
manrajgrover: Code quote types here as well

deeplab/README.md
```diff
@@ -0,0 +1,146 @@
...
+const getSemanticSegmentationMap = async (image) => {
+    return await model.segment(image)
+}
+```
+
+### Translating a Segmentation Map into the Color-Labelled Image
+
+To transform the segmentation map into a coloured image, use the `translate` method of the `SemanticSegmentation` object. Pass an optional canvas element as `canvas` to draw the output as a side effect.
+
+#### `model.translate` input
```
manrajgrover: I was thinking of calling `input` as `Params`, output as `Returns` and header for example as `Usage` and move away from `model.translate` since we explain it in `Usage`. Similar changes can be made above. What do you think?

deeplab/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+* **segmentationMap** :: RawSegmentationMap | tf.Tensor2D
```
manrajgrover: Code quotes here as well

deeplab/README.md
```diff
@@ -0,0 +1,146 @@
...
+## Contributing to the Demo
+
+Please see the demo [documentation](./demo/README.md).
+
+## Technical Details
+
+This model is based on the TensorFlow [implementation](https://github.com/tensorflow/models/tree/master/research/deeplab) of DeepLab v3. You might want to inspect the [conversion script](./convert_deeplab.sh), or download original pre-trained weights [here](https://github.com/tensorflow/models/blob/master/research/deeplab/g3doc/model_zoo.md). To convert the weights locally, run the script as follows, replacing `dist` with the target directory:
+
+```bash
+./utils/convert_deeplab.sh dist
```
manrajgrover: Since this is more of a script, let's call this folder `scripts` and make the change here as well

deeplab/demo/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "version": "0.0.1",
+  "description": "",
+  "main": "src/index.js",
+  "license": "Apache-2.0",
+  "private": true,
+  "engines": {
+    "node": ">=8.9.0"
+  },
+  "dependencies": {
+    "@tensorflow-models/deeplab": "link:.yalc/@tensorflow-models/deeplab",
```
manrajgrover: Commenting here to remember to change this after release is made.

deeplab/demo/src/index.js
```diff
@@ -0,0 +1,172 @@
...
+      deeplab[modelName].dispose();
+    }
+    const model = new SemanticSegmentation(modelName, state.isQuantized);
+    deeplab[modelName] = model;
+    const toggler = document.getElementById(`toggle-${modelName}-image`);
+    toggler.onclick = () => setImage(deeplabExampleImages[modelName]);
+    const runner = document.getElementById(`run-${modelName}`);
+    runner.onclick = async () => {
+      runner.classList.add('is-loading');
+      await sleep(100);
```
manrajgrover: Do we need to do this?

deeplab/demo/src/index.js
```diff
@@ -0,0 +1,172 @@
...
+};
+
+const status = message => {
+  const statusMessage = document.getElementById('status-message');
+  statusMessage.innerText = message;
+};
+
+const runDeeplab = async modelName => {
+  status(`Running the inference...`);
+  await sleep(100);
```
manrajgrover: Same here

deeplab/demo/src/index.js
```diff
@@ -0,0 +1,172 @@
...
+  const initialisationStart = performance.now();
+  const isQuantizationDisabled = document.getElementById(
+    'is-quantization-disabled'
+  ).checked;
+  if (!(isQuantizationDisabled ^ state.isQuantized)) {
+    state.isQuantized = !isQuantizationDisabled;
+    initialiseModels();
+    sleep(100);
```
manrajgrover: Same here

deeplab/demo/src/index.js
```diff
@@ -0,0 +1,172 @@
...
+  }
+  const input = document.getElementById('input-image');
+  if (!input.src || !input.src.length || input.src.length === 0) {
+    status('Failed! Please load an image first.');
+    return;
+  }
+
+  const model = deeplab[modelName];
+  if (input.complete && input.naturalHeight !== 0) {
```
manrajgrover: Seems like repetitive code here. Can we move prediction logic to separate function?

deeplab/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@tensorflow-models/deeplab",
+  "version": "0.1.0",
+  "description": "Model starter in TensorFlow.js",
```
manrajgrover: We should describe `deeplab` here

deeplab/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "module": "dist/deeplab.esm.js",
+  "unpkg": "dist/deeplab.min.js",
+  "jsdelivr": "dist/deeplab.min.js",
+  "types": "dist/index.d.ts",
+  "model": {
+    "name": "deeplab"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/sdll/tfjs-model-starter"
```
manrajgrover: We should change this:

```suggestion
    "url": "https://github.com/tensorflow/tfjs-models"
```

deeplab/src/deeplab_test.ts
```diff
@@ -0,0 +1,58 @@
...
+  NODE_ENVS,
+} from '@tensorflow/tfjs-core/dist/jasmine_util';
+import { SemanticSegmentation } from '.';
+
+describeWithFlags('SemanticSegmentation', NODE_ENVS, () => {
+  beforeAll(() => {
+    jasmine.DEFAULT_TIMEOUT_INTERVAL = 20000;
+  });
+
+  it('SemanticSegmentation predict method should not leak', async () => {
```
manrajgrover: We should check leaks in other operations also

deeplab/src/index.ts
```diff
@@ -0,0 +1,94 @@
...
+      const data = toInputTensor(input);
+      return tf.squeeze(model.execute(data) as tf.Tensor);
+    }) as RawSegmentationMap;
+  }
+
+  public async translate(
+    rawSegmentationMap: RawSegmentationMap,
+    canvas?: HTMLCanvasElement
+  ): Promise<SegmentationData> {
+    return processSegmentationMap(rawSegmentationMap, canvas);
```
manrajgrover: We should wrap this in `tf.tidy` also

deeplab/src/index.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  public async predict(
+    input: DeepLabInput,
+    canvas?: HTMLCanvasElement
+  ): Promise<DeepLabOutput> {
+    const rawSegmentationMap = await this.segment(input);
```
manrajgrover: Wrap this in `tf.tidy`?

deeplab/src/utils.ts
```diff
@@ -0,0 +1,106 @@
...
+  const labelNames = config['LABELS'];
+  const legend: Legend = Array.from(labels).reduce(
+    (accumulator, label) => ({
+      ...accumulator,
+      [labelNames[label]]: colormap[label],
+    }),
+    {}
+  );
+
+  return [legend, segmentationMap];
```
manrajgrover: Just thinking if it would be better to use `Object` here instead of an `Array`

deeplab/.prettierrc.js
```diff
@@ -0,0 +1,26 @@
+/**
```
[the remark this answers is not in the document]
sdll: I wonder how well this fits into the js/ts dev ecosystem, given the [substantially greater popularity](https://www.npmtrends.com/clang-format-vs-prettier) of prettier over clang-format. Prettier is more explicit: even[Google TypeScript guide](https://github.com/google/gts) has [switched](https://github.com/google/gts/issues/286) to Prettier. Given that the [tfjs](https://github.com/tensorflow/tfjs-core/blob/master/DEVELOPMENT.md) contribution guide was last updated before this change, I wonder whether it begs for an update. Let me know if this is a problem.

deeplab/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+* **width** :: number
+
+The new height of the processed image.
+
+* **SegmentationMap** :: Uint8ClampedArray
+
+The colored segmentation map as Uint8ClampedArray which can be [fed](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Pixel_manipulation_with_canvas) into ImageData and mapped to a canvas.
+
+#### Segmentation example
```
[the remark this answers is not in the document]
sdll: There is a`Usage` header on the top.

yalc.lock
```diff
@@ -0,0 +1,4 @@
+{
```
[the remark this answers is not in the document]
sdll: I wonder why the gitignore rule has missed this.

deeplab/demo/src/index.js
sdll: This allows the main thread to animate the loading state of the button. Let me know if there is a better way to do this.

deeplab/src/deeplab_test.ts
sdll: This test is time-intensive and will slow down the testing pipeline. The predict method executes all other segmentation subtasks. Is this really necessary?

deeplab/src/utils.ts
sdll: Agreed, I will change the output of the predict method as well.

deeplab/src/index.ts
sdll: `tf.tidy` cannot include async methods inside the lambda.

deeplab/src/index.ts
sdll: Ditto, tf.tidy cannot include async methods inside the lambda.

deeplab/src/utils.ts
```diff
@@ -0,0 +1,106 @@
...
+ * limitations under the License.
+ * =============================================================================
+ */
+
+export const createPascalColormap = (): Color[] => {
+  const colormap = new Array(config['DATASET_MAX_ENTRIES']['PASCAL']);
+  for (let idx = 0; idx < config['DATASET_MAX_ENTRIES']['PASCAL']; ++idx) {
+    colormap[idx] = new Array(3);
+  }
+  for (let shift = 7; shift > 4; --shift) {
```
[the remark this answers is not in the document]
sdll: This simply happens to be the colormap chosen by the PASCAL dev team. I have rewritten the [Python implementation](https://github.com/tensorflow/models/blob/master/research/deeplab/utils/get_dataset_colormap.py). I have just noticed that I did not produce the colormaps for Cityscapes and ADE20K, which resulted in that strange output I told you. I will ship a commit fixing this.

deeplab/README.md
sdll: I like that headers reflect what method we are talking about, which makes skimming the page easier.

sdll:
The CityScapes problem might be related to [this](https://github.com/tensorflow/models/issues/4525) issue.

deeplab/demo/src/index.js
manrajgrover: I think it can be solved using [`tf.nextFrame()`](https://js.tensorflow.org/api/latest/#nextFrame). Have you explored it?

deeplab/.prettierrc.js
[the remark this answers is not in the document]
manrajgrover: I think this is something that @dsmilkov @nsthorat can help with.

deeplab/src/deeplab_test.ts
manrajgrover: Since this test only checks for leaks and it does call other operations, we can skip tests for other operations.

deeplab/src/utils.ts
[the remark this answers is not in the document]
manrajgrover: In that case, we should reference their implementation here.

deeplab/demo/src/index.html
```diff
@@ -0,0 +1,188 @@
...
+          </div>
+
+        </div>
+      </div>
+
+    </div>
+  </div>
+</body>
+
+</html>
```
manrajgrover: Let's add a new line after every EOF

deeplab/demo/src/index.js
```diff
@@ -0,0 +1,178 @@
...
+  canvas.width = width;
+  canvas.height = height;
+  ctx.putImageData(segmentationMapData, 0, 0);
+
+  const legendList = document.getElementById('legend');
+  while (legendList.firstChild) {
+    legendList.removeChild(legendList.firstChild);
+  }
+
+  // Work around the parcel failure with Object.keys
```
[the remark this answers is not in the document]
sdll: `Uncaught (in promise) ReferenceError: _Object$keys is not defined`

Seems like the long tail of the transpiler issue which affects mangling as well.

deeplab/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+import * as tf from '@tensorflow/tfjs';
+import { config } from './config';
```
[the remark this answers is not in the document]
sdll: Could you please elaborate what the problem is, please? All files are formatted with prettier according to the config by default, might this be the config issue?

deeplab/src/utils_test.ts
```diff
@@ -0,0 +1,284 @@
+import { getColormap, translateLabels } from './utils';
+import { config } from './config';
+import { SemanticSegmentationBaseModel } from './types';
```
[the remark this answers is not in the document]
sdll: `tslint` shows no problems when I do `yarn lint`, and `prettier` does nothing new.

deeplab/src/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+const createADE20KColormap = (): Color[] => {
+  return config['COLORMAPS']['ADE20K'] as Color[];
+};
+
+const createCityscapesColormap = (): Color[] => {
+  return config['COLORMAPS']['CITYSCAPES'] as Color[];
+};
+
+export const getColormap = (base: SemanticSegmentationBaseModel) => {
+  if (base === 'pascal') {
```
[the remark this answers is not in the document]
sdll: This will break the modularity of the code and make it less maintainable, since   the pascal colormap is recreated, not read from the config. I prefer keeping this as a utility function.

deeplab/src/utils_test.ts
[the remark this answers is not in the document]
manrajgrover: Not sure why that is happening but take note here:

https://github.com/tensorflow/tfjs-models/blob/e590dedf45d093d67c2b31c05da7a407821c1629/deeplab/demo/src/index.js#L19

and the above lines have inconsistent spacing.

deeplab/src/index.ts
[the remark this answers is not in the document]
manrajgrover: See the comment below. I believe this is happening because of `prettier`.
Let's take input from Daniel and Nikhil on this one.

deeplab/demo/src/index.js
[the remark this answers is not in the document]
manrajgrover: That's strange. Thanks for adding a TODO here. It will help investigate this later on.

deeplab/demo/src/index.js
```diff
@@ -0,0 +1,180 @@
...
+const status = (message) => {
+  const statusMessage = document.getElementById('status-message');
+  statusMessage.innerText = message;
+};
+
+const runPrediction = (modelName, input, initialisationStart) => {
+  const model = deeplab[modelName];
+  model.predict(input).then((output) => {
+    displaySegmentationMap(modelName, output);
+    status(`Ran in ${performance.now() - initialisationStart} ms`);
```
manrajgrover: Let's round off this time to 2 floating places.

deeplab/demo/src/index.html
```diff
@@ -0,0 +1,185 @@
...
+                      </footer>
+                    </div>
+                  </div>
+                  <div class="column">
+                    <div class="card large">
+                      <div class="card-content">
+                        <div class="content">
+                          <b>DeepLab
+                            <a href="https://www.cityscapes-dataset.com">
+                              Cityscapes
```
manrajgrover: I was thinking of adding a warning here regarding its slowness.

deeplab/src/utils_test.ts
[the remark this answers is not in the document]
sdll: The demo eslint config is different from the tslint config. I will add a matching rule to eslint.

deeplab/src/config.ts
```diff
@@ -0,0 +1,384 @@
+export const config = {
+  BASE_PATH: 'https://storage.googleapis.com/gsoc-tfjs/models/deeplab',
```
startupgurukul: move to an official path away from gsoc-tfjs may be?

manrajgrover: Yes, that is something that Core team will do before publishing it. :)

sdll: @startupgurukul, there is also a [script](https://github.com/sdll/tfjs-models/blob/models/deeplab/deeplab/scripts/convert_deeplab.sh) to convert the weights.

kavikode:
please add script tag for DeepLab

sdll:
@kavikode, do you mean pulling DeepLab from npm and the like, which will happen after the PR is reviewed and merged? If so, you will be able to use the model simply by putting the following on your page:

```html
<script src="https://cdn.jsdelivr.net/npm/@tensorflow/tfjs"></script>
<script src="https://cdn.jsdelivr.net/npm/@tensorflow-models/deeplab"></script>
```

kavikode:
thanks so much @sdll that is what I meant

sdll:
@oveddan, I get the same issue when switching the text detection model to the pattern you have described, see [here](https://github.com/tensorflow/tfjs-models/blob/6b89db26f7daf72d62d58bf66a02a9edfb902881/text-detection/src/index.ts#L45). There seems to be an issue with loading via `@tensorflow/tfjs-converter`, since everything works well when loading via `@tensorflow/tfjs`.

sdll:
@oveddan, I have worked around the issue by pinning the `tfjs` version to 1.2.2 for now.

deeplab/README.md
```diff
@@ -0,0 +1,202 @@
...
+import {load} from '@tensorflow-models/deeplab';
+const modelName = 'pascal';   // set to your preferred model, out of `pascal`, `cityscapes` and `ade20k`
+const quantizationBytes = 2; // either 1, 2 or 4
+const url = 'https://storage.googleapis.com/gsoc-tfjs/models/deeplab/quantized/1/pascal/model.json';
+const modelPromise = load({modelUrl: url});
+```
+
+This will initialize the model and return the promise of a `SemanticSegmentation` object.
+
+You can still pass the `base` attribute set to either `pascal`,`pascal`, `cityscapes` and `ade20k` in the argument to use the corresponding colormap and labelling scheme. Otherwise, you would have to provide those yourself.
```
manrajgrover:
```suggestion
You can set the `base` argument to `pascal`, `cityscapes` or `ade20k` to use the corresponding colormap and labelling scheme. Otherwise, you would have to provide those yourself during segmentation.
```

deeplab/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+If you require more careful control over the initialization and behavior of the model (e.g. you want to use your own labelling scheme and colormap), use the `SemanticSegmentation` object on its own, passing a pre-loaded `GraphModel` in the constructor:
```
manrajgrover:
```suggestion
If you require more careful control over the initialization and behavior of the model (e.g. you want to use your own labelling scheme and colormap), use the `SemanticSegmentation` class passing a pre-loaded `GraphModel` in the constructor:
```

deeplab/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+```typescript
+import * as tfconv from '@tensorflow/tfjs-converter';
+import {SemanticSegmentation, getURL} from '@tensorflow-models/deeplab';
+// use the getURL utility function to get the URL to the pre-trained weights
+const base = 'pascal';   // set to your preferred model, out of `pascal`, `cityscapes` and `ade20k`
+const quantizationBytes = 2; // either 1, 2 or 4
+const modelUrl = getURL(base, quantizationBytes);
+const rawModel = tfconv.loadGraphModel(modelUrl);
```
manrajgrover:
```suggestion
const graphModel = await tfconv.loadGraphModel(modelUrl);
```

deeplab/README.md
```diff
@@ -0,0 +1,202 @@
...
+const modelPromise = (async () => {return new SemanticSegmentation(await rawModel)();
```
manrajgrover:
```suggestion
const model = new SemanticSegmentation(graphModel);
```

manrajgrover: You would need to pass `base` or give a default value otherwise `model.segment` will fail since `this.base` will be `undefined` and `getColormap` and `getLabels` will throw an error. I think there are couple of ways to move forward:

1. Either we don't give user an option to provide their own url like other tfjs-models. I think this is the best way forward and simplifies usage a lot.
2. Check all flows where code might break and fix them.

I'll say let's stick with 1st point.

deeplab/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+#### `model.segment` input
+
+- **image** :: `ImageData | HTMLImageElement | HTMLCanvasElement | HTMLVideoElement | tf.Tensor3D`;
+
+The image to segment
+
+- **canvas** (optional) :: `HTMLCanvasElement`
+
+Pass an optional canvas element as `canvas` to draw the output as a side effect.
```
manrajgrover:
```suggestion
Pass an optional canvas element as `canvas` to draw the output.
```

deeplab/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+The array of names corresponding to labels
+
+- **segmentationMap** :: `tf.Tensor2D`
+
+The segmentation map of the image
+
+- **canvas** (optional) :: `HTMLCanvasElement`
+
+Pass an optional canvas element as `canvas` to draw the output as a side effect.
```
manrajgrover:
```suggestion
Pass an optional canvas element as `canvas` to draw the output
```

deeplab/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+```javascript
+const getSemanticSegmentationMap = async (image) => {
+    return await model.predict(image)
+}
+```
+
+### Translating a Segmentation Map into the Color-Labelled Image
+
+To transform the segmentation map into a coloured image, use the `toSegmentationImage` method of the `SemanticSegmentation` object.
```
manrajgrover: `SemanticSegmentation` does not have a `toSegmentationImage` method and are part of `utils`.

deeplab/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+#### `model.toSegmentationImage` input
```
manrajgrover: This should not be coming from `model` and likewise below

deeplab/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+#### `model.predict` input
+
+- **image** :: `ImageData | HTMLImageElement | HTMLCanvasElement | HTMLVideoElement | tf.Tensor3D`;
+
+The image to segment
+
+#### `model.predict` output
+
+- **rawSegmentationMap** :: `Promise<tf.Tensor2D>`
```
manrajgrover: I don't think this method returns a `Promise`. Kindly check again.

deeplab/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+The segmentation map of the image
+
+- **canvas** (optional) :: `HTMLCanvasElement`
+
+Pass an optional canvas element as `canvas` to draw the output as a side effect.
+
+#### `model.toSegmentationImage` output
+
+The returned promise is the future of a `SegmentationData` object that contains two attributes:
```
manrajgrover:
```suggestion
A promise resolving to `SegmentationData` object that contains two attributes:
```

deeplab/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+- **segmentationMap** :: `Uint8ClampedArray`
+
+The colored segmentation map as `Uint8ClampedArray` which can be [fed](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Pixel_manipulation_with_canvas) into `ImageData` and mapped to a canvas.
+
+#### `model.toSegmentationImage` example
+
+```javascript
+const translateSegmentationMap = async (segmentationMap) => {
+    return await model.toSegmentationImage(segmentationMap)
```
manrajgrover: `toSegmentationImage` takes more than one params as defined above.

deeplab/demo/src/index.css
```diff
@@ -0,0 +1,19 @@
+.is-on-top {
+  z-index: 100;
+}
+.is-flex-column {
```
manrajgrover:
```suggestion

.is-flex-column {
```

deeplab/demo/src/index.css
```diff
@@ -0,0 +1,19 @@
...
+  display: block;
+}
+
+.is-vcentered {
+  align-items: center;
+  margin-top: auto;
+  margin-bottom: auto;
+}
+
+.m-r-sm { margin-right: 0.5rem; }
```
manrajgrover: Please move this to separate lines

deeplab/demo/src/index.html
```diff
@@ -0,0 +1,211 @@
...
+                    <option>1</option>
+                    <option selected>2</option>
+                    <option>4</option>
+                  </select>
+                </div>
+                <span class="icon is-left">
+                  <i class="fas fa-globe"></i>
+                </span>
+              </div>
+              <!-- <b>Disable quantization</b>:
```
manrajgrover: Let's clean up the code if not needed

deeplab/rollup.config.js
```diff
@@ -0,0 +1,108 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+import resolve from 'rollup-plugin-node-resolve';
+import {terser} from 'rollup-plugin-terser';
+import typescript from 'rollup-plugin-typescript2';
+// import {uglify} from 'rollup-plugin-uglify';
```
manrajgrover: Lets clean up here

deeplab/rollup.config.js
```diff
@@ -0,0 +1,108 @@
...
+    * distributed under the License is distributed on an "AS IS" BASIS,
+    * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    * See the License for the specific language governing permissions and
+    * limitations under the License.
+    * =============================================================================
+    */`
+};
+
+function minify() {
+  // return uglify({
```
manrajgrover: Let's clean up here as well

deeplab/scripts/convert_deeplab.sh
```diff
@@ -0,0 +1,248 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =============================================================================
+
+# Make sure that we run Python 3.6, not 3.7
+PYTHON_VERSION=$(python -V 2>&1 | sed 's/.* \([0-9]\).\([0-9]\).*/\1\2/')
+if [ "$PYTHON_VERSION" -gt "36" ]; then
+  echo "This script requires python 3.6 or older."
+  # exit 1
```
manrajgrover: Let's uncomment this.
```suggestion
  exit 1
```

deeplab/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+  public predict(input: DeepLabInput): tf.Tensor2D {
+    return tf.tidy(() => {
+      const data = toInputTensor(input);
+      return tf.squeeze(this.model.execute(data) as tf.Tensor);
+    }) as tf.Tensor2D;
+  }
+
+  public async segment(
+      input: DeepLabInput, canvas?: HTMLCanvasElement,
+      colormap = getColormap(this.base),
```
manrajgrover: As explained before, this will break where `this.base` is not defined

kavikode:
can't wait! Thank you!

kavikode:
Can you please include an example how we can use transfer learning to retrain the DeepLab model for custom data objects in the browser? I'd like to perform this transfer learning with the browser in real-time. How can we  freeze the layers to train only the last couple of layers in the DeepLab model? 

I understand this is possible in Python. I also understand that it was possible to retrain the models within the browser for mobilenet, speech commands, and coco-ssd within the browser using Tensorflow.js. I imagine that it is possible to do the same retraining with the Deeplab model within the browser. Please help. @sdll @oveddan @manrajgrover @tylerzhu-github @tafsiri

sdll:
@kavikode,since this implementation uses TF FrozenModels exported to TF.js GraphModels as the backbone, it cannot be trained in the browser. You can, however, train your own weights with Python (the code is [here](https://github.com/tensorflow/models/tree/master/research/deeplab), some instructions are in the [FAQ](https://github.com/tensorflow/models/blob/master/research/deeplab/g3doc/faq.md)), and then convert the frozen model using this [script](https://github.com/sdll/tfjs-models/blob/models/deeplab/deeplab/scripts/convert_deeplab.sh).

kavikode:
Thank you so much @sdll :-) I will following your guide.

Likewise, will it be possible to work on live camera feed in real-time with 30 frames per seconds with this model?

I did find https://intel.github.io/webml-polyfill/examples/semantic_segmentation/?prefer=none&b=WebGL&m=deeplab_mobilenet_v2_224_tflite&s=camera&d=0

The frame rate is too low.

Can you please include a demo of live camera?

sdll:
@kavikode, not sure that we can get up to 30 FPS, as the screenshots above show,  < 3 FPS is the norm, which is not that far away from the demo you have mentioned.

kavikode:
@sdll thanks Is there a way to perform optimization techniques because I believe that in Python it ran 30 fps?

Do you know of any other segmentation models that would run in 30 FPS real-time in Tensorflow.js?

sdll:
> @sdll thanks Is there a way to perform optimization techniques because I believe that in Python it ran 30 fps?

I will look into it when I have more bandwidth. If you find anything by then, let me know!

> Do you know of any other segmentation models that would run in 30 FPS real-time in Tensorflow.js?

No, unfortunately

PR #231, opened 2019-06-06
update dependencies to address security warnings

tafsiri:
context: https://github.com/tensorflow/tfjs-models/network/alerts

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/231)


body-pix/demos/package.json
```diff
@@ -48,4 +48,4 @@
   "eslintIgnore": [
     "dist/"
   ]
-}
+}
```
caisq: nit: Retain the ending line break.

tafsiri: Done

PR #235, opened 2019-06-11
allow user to provide a custom url for the mobilenet model

pyu10055:
added an optional parameter for specifying the custom model url.
updated the version for mobilenet.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/235)


tafsiri:
@pyu10055 FYI some tests seem to be failing still.

tafsiri:
@pyu10055 it looks like the signature of the constructor changed when this merged. Could we have it match the signature of the load function since (i.e. take the config object) since it is something we export and people can use directly.

PR #237, opened 2019-06-16
update models peerDependencies to be individual npm instead of union

pyu10055:
This is to reduce the peerDependencies to be the tfjs sub packages that are used in the model.
This would reduce the bundle size for applications that use the model npms. 
toxicity model is not updated, since it has deps on universal sentence encoder.
speech-commands is not updated since it uses all packages.

updating tfjs related npm version to the latest, and minor version update for all involved models.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/237)


nsthorat:
Thanks Ping! Can you take care of publishing these models as well? Thanks!

PR #241, opened 2019-06-18
updated posenet readme to reflect on inference params being an object, and solidify config object typings.

oveddan:
Fixed the posenet readme to have inference params being an object.  Also fixes the naming of maxDetections parameter in the readme.

Additionally, adds better typing to the inference config objects, and treats them as immutable when setting defaults.

 I'm finding the consolidation of estimateMultiplePoses and estimateSinglePose into a single api method a bit confusing for a few reasons:

1.  It's a semi-big change from before which took my head a bit of time to wrap around, and I think its going to be confusing to people that are used to the existing api.
2. The options object has different paremeters depending on the estimation method, which is a bit tough to document - it was challenging to explaining what the parameter `decodingMethod` means.  why require the user to specify "decodingMethod" to be "single-person" or "multi-person" instead of just having two different api methods with parameters that are specific to those methods?  

It may be too late to change this, but I think it would be cleaner and easier to consume as two separate api methods.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/241)

### DIFF
--- a/posenet/README.md
+++ b/posenet/README.md
@@ -131,7 +131,9 @@ const pose = poses[0];
 
 * **image** - ImageData|HTMLImageElement|HTMLCanvasElement|HTMLVideoElement
    The input image to feed through the network.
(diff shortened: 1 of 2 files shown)


oveddan:
Changes encapsulated in #242

PR #242, opened 2019-06-19
PoseNet - Split pose estimation methods  into estimateSinglePose and estimateMultiplePoses, and move inputResolution specification from loading to inference time.

oveddan:
PoseNet - to make a cleaner api, split the existing pose estimation method into estimateSinglePose and estimateMultiplePoses with an options object with parameters specific to that object.

Moved inputImageResolution into the inference params, since it's not needed at loaded time.  In demos, no longer re-loading the model when inputImageResolution is changed.

* Fixed documentation that had out of date input image resolution sizes.
* Consolidated functionality in estimation methods into encaspulated functions that are portable and reusable.

* Solidify typings of inference config objects. 
* Treating the config objects as immutable.
* Fixed the posenet readme to have inference params being an object. 
* Also fixes the naming of maxDetections parameter in the readme.

Additionally, adds better typing to the inference config objects, and treats them as immutable when setting defaults.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/242)

### DIFF
--- a/body-pix/demos/package.json
+++ b/body-pix/demos/package.json
@@ -9,8 +9,8 @@
     "node": ">=8.9.0"
   },
   "dependencies": {
-    "@tensorflow-models/body-pix": "1.0.1",
-    "@tensorflow/tfjs": "1.1.0",
+    "@tensorflow-models/body-pix": "1.1.0",
+    "@tensorflow/tfjs": "1.2.1",
     "stats.js": "0.17.0"
   },
   "scripts": {
--- a/body-pix/demos/yarn.lock
+++ b/body-pix/demos/yarn.lock
@@ -755,10 +755,8 @@
     "@parcel/utils" "^1.11.0"
     physical-cpu-count "^2.0.0"
 
-"@tensorflow-models/body-pix@1.0.1":
-  version "1.0.1"
-  resolved "https://registry.yarnpkg.com/@tensorflow-models/body-pix/-/body-pix-1.0.1.tgz#bcf0b680b14dd65f7249a922d16a5afb132c1549"
-  integrity sha512-s63eE+ns6ArGZ6MiX7t95yzUPUxdUkEUFwG8jWkhAlDSANiqFFM1rmNcIp2iF2ldojftFRr2y7vwHGocFYrDdQ==
+"@tensorflow-models/body-pix@file:.yalc/@tensorflow-models/body-pix":
+  version "1.0.1-5c8afa55"
 
 "@tensorflow/tfjs-converter@1.1.0":
   version "1.1.0"
--- a/body-pix/package.json
+++ b/body-pix/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow-models/body-pix",
-  "version": "1.0.1",
+  "version": "1.1.0",
   "description": "Pretrained BodyPix model in TensorFlow.js",
   "main": "dist/index.js",
   "jsnext:main": "dist/body-pix.esm.js",
@@ -13,10 +13,12 @@
     "url": "https://github.com/tensorflow/tfjs-models.git"
   },
   "peerDependencies": {
-    "@tensorflow/tfjs": "^1.1.0"
+    "@tensorflow/tfjs-core": "~1.2.1",
+    "@tensorflow/tfjs-converter": "~1.2.1"
   },
   "devDependencies": {
-    "@tensorflow/tfjs": "^1.1.0",
+    "@tensorflow/tfjs-core": "~1.2.1",
+    "@tensorflow/tfjs-converter": "~1.2.1",
     "@types/jasmine": "~2.5.53",
     "jasmine": "~3.2.0",
     "jasmine-core": "~3.1.0",
--- a/body-pix/src/body_pix_model.ts
+++ b/body-pix/src/body_pix_model.ts
@@ -16,7 +16,8 @@
  * =============================================================================
  */
 
-import * as tf from '@tensorflow/tfjs';
+import * as tfconv from '@tensorflow/tfjs-converter';
+import * as tf from '@tensorflow/tfjs-core';
 
 import {checkpoints} from './checkpoints';
 import {decodePartSegmentation, toMask} from './decode_part_map';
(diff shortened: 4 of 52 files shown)


nsthorat:
LGTM modulo bringing back the old method (it can just forward to the new methods).

Thanks for taking this!

tylerzhu-github:
LGTM. One minor comment is that the demo now replies on the updated API, and thus we might need to release a new NPM and update the demo's PoseNet dependency to the latest one.

tylerzhu-github:
I prefer it to be specified during loadTime. This makes input resolution validation upfront 

> Also - even though inputResolution was moved in this pr to inference time, I wouldn't mind it being instead specified at loadTime, how it was before this change. The advantage of it being at load time is that all the performance configuration options are together in one place. What do you all think?
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/242#-:-LhkFSMq-fnVeuf8CDxU:bu3gy1w)_ status: :shipit: complete! 1 of 1 approvals obtained (waiting on @oveddan and @tylerzhu-github)

tylerzhu-github:
The benefit of validate the input resolution only once on the loadTime is to avoid duplicated validation for each inference. Thanks for the through rethinking about the best class to hold the input resolution variable. I agree that PoseNet class would be a better class to store the input resolution variable.

> Just curious, what's the advantage of doing input resolution validation upfront?
> 
> Also, if we are to specify it at load time, it doesn't make sense for the BaseModel class to contain the inputResolution property when it's not actually used in that class. The PoseNet class is the one that uses that property during inference, so it belongs either as a property of the PoseNet model or as an argument to the inference methods.
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/242#-:-LhlxkeoEGfuR7ZkdgLK:b79cen4)_ status: :shipit: complete! 1 of 1 approvals obtained (waiting on @tylerzhu-github)
> 
> _[posenet/README.md, line 102 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/242#-Lhk6Rv23Oip4c1F7TiS:-LhkFlt99WRaxMjvezcQ:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/a22e47f99a874669f29c407e22b0d03a5a7b2ac3/posenet/README.md#L102)):_
> 
> Previously, tylerzhu-github (tylerzhu-github) wrote…
> Done.
> 
> _[posenet/README.md, line 163 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/242#-Lhk6bbW2j5HnEAFkwq6:-LhkFmLR4OPe1Bz63PDV:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/a22e47f99a874669f29c407e22b0d03a5a7b2ac3/posenet/README.md#L163)):_
> 
> Previously, oveddan (Dan Oved) wrote…
> Done.
> 
> _[posenet/demos/camera.js, line 350 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/242#-Lhk7xBa0mSxADORGV2L:-LhkFnGHB485QwFZZBK8:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/a22e47f99a874669f29c407e22b0d03a5a7b2ac3/posenet/demos/camera.js#L350)):_
> 
> Previously, oveddan (Dan Oved) wrote…
> Done.

tylerzhu-github:
> I totally understand the need to avoid duplicating validation. I was concerned with this in the original release of PoseNet where we were doing validation of many parameters on every inference call. However, after doing some performance tests these validation calls took pretty much no time at all; the bottleneck in performance was all in the feed forward step.
> 
> Also, according to this logic of moving inputResolution back to the loadConfig to avoid repeat validation, we would also have to move all of the options from InferenceConfig into the load config, such as nsmradius and and max detections. This wouldn't make much sense since they are not needed for loading.
> 
> Since inputResolution is no longer needed for loading, I think it's better for clarity purposes to only take it when its used, and also it gives the user the ability to change the input image size without needing to reload the model, which is particularly useful when someone wants to try a bunch of different sizes to get the right performance when using the resnet model.
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/242#-:-LhoRUh07hMH4YQZuWCc:b-88ydjw)_ status: :shipit: complete! 1 of 1 approvals obtained (waiting on @tylerzhu-github)

The inference config contains all the parameters that is needed for post-processing on the predicted tensors
.
The loadConfig declares all the information needed for the model to perform a forward pass. 

Maybe a compromised way is to have 1 additional method called `setInputResolution` so that a user can config the PoseNet input resolution on-the-fly?

tylerzhu-github:
Or we can bring back the previous scale factor from 1.0 and implicitly compute the closest valid resolution. This will remove the need to store the inputResolution variable.

oveddan:
___
*[posenet/package.json, line 3 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/242#-LhvLQmm1aKB8YRYWL4u:-LhvMLKK1VDzOUA7dXmT:bgt7vwa) ([raw file](https://github.com/tensorflow/tfjs-models/blob/a1e19e478a87f3c26e70d308bf7b26f4ed4ee0d8/posenet/package.json#L3)):*
<details><summary><i>Previously, tylerzhu-github (tylerzhu-github) wrote…</i></summary><blockquote>

> ```
> 2.1.0
> ```
The demo now uses the new API introduced in the PR, I wonder if we can still make the demo backward compatible to support both 2.0.0 and 2.1.0 API.
</blockquote></details>

It should - estimatePoses was added back, and that just calls the new api.

oveddan:
___
*[posenet/README.md, line 100 at r5](https://reviewable.io/reviews/tensorflow/tfjs-models/242#-LhvNb4y-IgJRRcxngol:-LhvO4Ld4ana7iNzee1o:b-6pn1wd) ([raw file](https://github.com/tensorflow/tfjs-models/blob/bdc34433c44b050eaf578133e4150a781d056c56/posenet/README.md#L100)):*
<details><summary><i>Previously, tylerzhu-github (tylerzhu-github) wrote…</i></summary><blockquote>

It's odd for the estimateSinglePose(..) to return an array of Poses. Should it be just one Pose object?
</blockquote></details>

bug in documentation.  it's been fixed.

nsthorat:
LGTM, can you merge and fix the conflicts?

PR #252, opened 2019-07-02
norm control

bileschi:
Any thoughts on how / whether to test this?

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/252)


mobilenet/src/index.ts
```diff
@@ -115,20 +143,31 @@ export interface MobileNet {
 class MobileNetImpl implements MobileNet {
   model: tfconv.GraphModel;
 
+  // Values read from images are in the range [0.0, 255.0], but they must
+  // be normalized to [min, max] before passing to the mobilenet classifier.
+  // Different implementations of mobilenet have different values of [min, max].
+  // We store the appropriate normalization parameters using these two scalars
+  // such that:
+  // out = in * normalizationScale + normalizationOffset;
+  private normalizationScale: tf.Scalar;
```
nsthorat: these don't need to be scalars anymore since we can just take numbers

mobilenet/src/index.ts
```diff
@@ -47,24 +47,52 @@ const EMBEDDING_NODES: {[version: string]: string} = {
   '2.00': 'module_apply_default/MobilenetV2/Logits/AvgPool'
 };
 
-const MODEL_INFO: {[version: string]: {[alpha: string]: string}} = {
+export interface MobileNetInfo {
+  // Where to find the TFHub version of this model.
+  TFHubURL: string;
```
nsthorat: how about just url?

mobilenet/src/index.ts
```diff
@@ -115,20 +143,31 @@ export interface MobileNet {
...
+  // out = in * normalizationScale + normalizationOffset;
+  private normalizationScale: tf.Scalar;
   private normalizationOffset: tf.Scalar;
 
   constructor(
       public version: string, public alpha: string,
-      public modelUrl: string|tf.io.IOHandler) {
-    this.normalizationOffset = tf.scalar(127.5);
-  }
+      public modelUrl: string|tf.io.IOHandler) {}
```
nsthorat: lets add normalization bounds here and add it to the public load method

PR #254, opened 2019-07-06
landscape VS portrait webcam detection

pinch44:
I try to use posenet with my webcam in portrait mode . But detection is wrong.
I need to rotate 90 degrees. 
flipHorizontal is not enough to be concluant.

How can I do in setup function :
`function setup() {
    createCanvas(640, 480);
    video = createCapture(VIDEO);
    //video._translate(width,0);
    //angleMode(DEGREES);
    //video._rotate(-90);
    //video.hide();
    poseNet = ml5.poseNet(video, modelReady);
    poseNet.on('pose', gotPoses);
}`

I can rotate my webcam after detection in draw function but it is too late :
`function draw() {
    image(video, 0, 0);
    video._translate(640,0);
    angleMode(DEGREES);
    video._rotate(-90);
    [...]
}`

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/254)


tafsiri:
Hi, if this PR is still needed I think it would need to be remade correctly. It currently tries to merge into the fix_posenet branch but it should be trying to merge into master. So the diff is probably the inverse of what it should be.

PR #274, opened 2019-07-15
Add Text Detection

aptlin:
This PR adds a text detection model based on [PSENet](https://github.com/liuheng92/tensorflow_PSENet).

image

Here is the to-do list:

- [x] improve the performance by varying the cropping size and number of inferred kernels
- [x] replace OpenCV.js with vanilla TypeScript
- [x] make a proper demo
- [x] expand the docs
- [x] write tests for algos replacing OpenCV.js
- [x] refactor the Docker setup for end-to-end OpenCV.js build

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/274)


jenperson:
Thank you for your contribution! I'm very excited about this use case, but since it is not owned by Google, it can't be directly added to the models. Please consider adding a link to your code in the [Gallery](https://github.com/tensorflow/tfjs/blob/master/GALLERY.md).

PR #275, opened 2019-07-17
exposing body-pix tensor based ops

oveddan:
In many of my personal projects I find myself wanting to work with the tensors returned from bodypix, instead of the cpu based output.    This is particularly the case when wanting to do shader based ops on the outputs.

This PR exposes tensor based ops on the `BodyPix` class.

The two new API methods are:
`estimatePersonSegmentationOp`
`estimatePartSegmentationOp`

They both take 3d tensors as inputs and return 2d tensors.

The existing API methods are unchanged.  They wrap the new api methods and convert the output to be cpu based arrays.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/275)

### DIFF
--- a/body-pix/src/body_pix_model.ts
+++ b/body-pix/src/body_pix_model.ts
@@ -24,7 +24,7 @@ import {decodePartSegmentation, toMask} from './decode_part_map';
 import {assertValidOutputStride, MobileNet, MobileNetMultiplier, OutputStride} from './mobilenet';
 import {ModelWeights} from './model_weights';
 import {BodyPixInput, PartSegmentation, PersonSegmentation} from './types';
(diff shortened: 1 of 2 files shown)


nsthorat:
I wonder if estimatePersonSegmentation should take a config object like we do for posenet and then we should have a bit there that returns the activation?

nsthorat:
Hm yeah that's annoying. How about estimatePersonSegmentationActivation or something along those lines?

oveddan:
I signed it.

oveddan:
@nsthorat just wanted to bump this ;) I have some cool glitch sketches I want to publish that would use this.

nsthorat:
One small question then LGTM

PR #293, opened 2019-09-18
Posenet - relax input resolution requirement

oveddan:
This PR modifies the `posenet.load` function to allow any input resolution, and internally adjusts the input resolution to one that has its corresponding odd value be divisible by the output stride.

This makes the API easier to consume and easier to understand.

* Some "invalid" input resolutions were added to the demo as a form of test;  internally those invalid resolutions will be adjusted to be valid resolutions.

Added some tests around validation of the input resolution and setting it when the model is loaded.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/293)


oveddan:
Closing, because this is now included in #296 which allows for the input width and height to be specified.

PR #295, opened 2019-09-23
Extracting promise values from TensorBuffer function

esouthren:
Running the model was causing an 'incomplete alias for namespace' error due to this promise. 

Extracting the promise values and assigning them all directly seems to solve the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/295)


dsmilkov:
Fixed in  tensorflow/tfjs-models/pull/292/files thus closing this PR. If that is not the case, we can reopen

PR #299, opened 2019-09-26
simplifies and speeds up USE trie

1wheel:
This removes recursive branching and caches array creation.  

```
import {loadTokenizer} from './index'

loadTokenizer().then(tokenizer => {

  var gettysburg = 'Fourscore and seven years ago our fathers brought forth, on this continent, a new nation, conceived in liberty, and dedicated to the proposition that all men are created equal. Now we are engaged in a great civil war, testing whether that nation, or any nation so conceived, and so dedicated, can long endure. We are met on a great battle-field of that war. We have come to dedicate a portion of that field, as a final resting-place for those who here gave their lives, that that nation might live. It is altogether fitting and proper that we should do this. But, in a larger sense, we cannot dedicate, we cannot consecrate—we cannot hallow—this ground. The brave men, living and dead, who struggled here, have consecrated it far above our poor power to add or detract. The world will little note, nor long remember what we say here, but it can never forget what they did here. It is for us the living, rather, to be dedicated here to the unfinished work which they who fought here have thus far so nobly advanced. It is rather for us to be here dedicated to the great task remaining before us—that from these honored dead we take increased devotion to that cause for which they here gave the last full measure of devotion—that we here highly resolve that these dead shall not have died in vain—that this nation, under God, shall have a new birth of freedom, and that government of the people, by the people, for the people, shall not perish from the earth.'

  console.time('gettysburg')
  var tokens = tokenizer.encode(gettysburg)
  console.timeEnd('gettysburg') // 1650 ms -> 17 ms!
});
```

I also tried a simple hash, but at 40 ms it is slower than the trie; performance is closely tied to the longest token in the vocab. 

```
module.exports = function(){
  const hash = {}
  let longestToken = 0

  function insert(word, score, index){
    const symbols = []
    for (const symbol of word) {
      symbols.push(symbol)
    }

    hash[symbols] = [symbols, score, index]
    longestToken = Math.max(longestToken, symbols.length)
  }

  function commonPrefixSearch(symbols){
    const rv = symbols
      .slice(0, longestToken + 1)
      .map((d, i) => hash[symbols.slice(0, i + 1)])
      .filter(d => d)

    return rv.length ? rv : [[symbols[0]], 0, 0]
  }

  return {insert, commonPrefixSearch}
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/299)

### DIFF
--- a/universal-sentence-encoder/src/tokenizer/trie.ts
+++ b/universal-sentence-encoder/src/tokenizer/trie.ts
@@ -18,15 +18,13 @@
 import {stringToChars} from '../util';
 
 // [token, score, index]
-// TODO fix typing
-// type OutputNode = [string[], number, number];
-type OutputNode = any; 
+type OutputNode = [string[], number, number];
 
 class TrieNode {
   public parent: TrieNode;
   public end: boolean;
   public children: {[firstSymbol: string]: TrieNode};
-  public word: OutputNode[];
+  public word: OutputNode;
 
   constructor() {
     this.parent = null;
@@ -77,11 +75,11 @@ export class Trie {
     let node = this.root.children[ss[0]];
 
     for (let i = 0; i < ss.length && node; i++){
-      if (node.end) output.push(node.word);
+      if (node.end){ output.push(node.word); }
       node = node.children[ss[i + 1]];
     }
 
-    if (!output.length) output.push([[ss[0]], 0, 0]);
+    if (!output.length){ output.push([[ss[0]], 0, 0]); }
 
     return output;
   }
--- a/universal-sentence-encoder/src/tokenizer/trie.ts
+++ b/universal-sentence-encoder/src/tokenizer/trie.ts
@@ -18,39 +18,21 @@
 import {stringToChars} from '../util';
 
 // [token, score, index]
-type OutputNode = [string[], number, number];
+// TODO fix typing
+// type OutputNode = [string[], number, number];
+type OutputNode = any; 
 
 class TrieNode {
   public parent: TrieNode;
   public end: boolean;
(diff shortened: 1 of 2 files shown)


universal-sentence-encoder/src/tokenizer/trie.ts
```diff
@@ -18,67 +18,29 @@
 import {stringToChars} from '../util';
 
 // [token, score, index]
-type OutputNode = [string[], number, number];
+// TODO fix typing
+// type OutputNode = [string[], number, number];
+type OutputNode = any;
```
annxingyuan: Why doesn't the original typing work?

annxingyuan:
Thank you @1wheel ! Let's just get rid of the `any` and we're good to go!

universal-sentence-encoder/src/tokenizer/trie.ts
1wheel: Typescript complains that `this.word = [[], 0, 0];` doesn't have types; I'm not sure how to add them.

```
src/tokenizer/trie.ts:35:18 - error TS2739: Type '[]' is missing the following properties from type '[string[], number, number]': 0, 1, 2

35     this.word = [[], 0, 0];
```

annxingyuan: You need to change `public word: OutputNode[];` (line 29) to `public word: OutputNode;`

Looks like there are also two lint issues on lines 80, 84 - I think you just need to add curly braces around the if statements.

1wheel: Ah, that makes sense.

PR #303, opened 2019-10-07
body-pic demo update parcel bundler in yarn.lock

tafsiri:
Resolves security warning

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/303)


tafsiri:
@dsmilkov yep, the yarn lock just needs to update to the latest matching version (if you installed it without a yarn.lock you would also get the fix.)

PR #305, opened 2019-10-10
Rename bodypix methods to shorter names.

nsthorat:
The four methods look like:

- segment(Multi)Person
- segment(Multi)PersonParts

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/305)

### DIFF
--- a/body-pix/README.md
+++ b/body-pix/README.md
@@ -150,7 +150,7 @@ around a person but may result in some pixels being that are part of a person be
 
 #### Returns
 
-It returns a `Promise` that resolves with **an array** of `PersonSegmentation`s. When there are multiple people in the image, each `PersonSegmentation` object in the array represents one person. More details about the `PersonSegmentation` object can be found in the documentation of the `segmentSinglePerson` method.
+It returns a `Promise` that resolves with **an array** of `PersonSegmentation`s. When there are multiple people in the image, each `PersonSegmentation` object in the array represents one person. More details about the `PersonSegmentation` object can be found in the documentation of the `segmentPerson` method.
 
 
 ```javascript
@@ -292,7 +292,7 @@ around a person but may result in some pixels being that are part of a person be
 
 #### Returns
 
-It returns a `Promise` that resolves with **an array** of `PartSegmentation`s. When there are multiple people in the image, each `PartSegmentation` object in the array represents one person. More details about the `PartSegmentation` object can be found in the documentation of the `segmentSinglePersonParts` method.
+It returns a `Promise` that resolves with **an array** of `PartSegmentation`s. When there are multiple people in the image, each `PartSegmentation` object in the array represents one person. More details about the `PartSegmentation` object can be found in the documentation of the `segmentPersonParts` method.
 
 
 ```javascript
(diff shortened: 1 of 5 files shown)


oveddan:
@nsthorat agreed shorter names would be great.

One thing regarding the term 'single' - the 'single' methods actually provided masks for everyone in the image.  The only difference between the `single` and `multi` methods is that the `multi` methods provide **instance level** segmentations.  So i think it would be slightly misleading to name those methods `single` as they actually work for multiple people.  

One other thing to note is that even though the `single` methods you are speaking of return a mask that covers multiple people, it returns only a single pose.  I can see a common use case where people want a mask for all people in the image (non-instance level), but not a common one where they would want that mask plus a single pose.  So we may want to consider having something that doesnt return the single pose as to avoid confusion.

oveddan:
I for one, as a consumer of this API, wanted to just get the mask for all the people in the image (not instance level) and due to the method name of estimateSinglePerson thought that it wouldn't work with multiple people in the image, but was pleased to learn after digging into the code that it does actually work with multiple people, but just returns a single mask and pose.

oveddan:
Sorry for the comment spam - I think something like:
`segmentPerson` `segmentMultiPerson`
`segmentPersonParts` `segmentMultiPersonParts`

would be the clearest

tylerzhu-github:
How about we just have segmentMultiPerson and segmentMultiPersonPart and then allows one to chose whether we use instance or semantic decoding? This will simplify the API drastically with only 2 method.

oveddan:
@tylerzhu-github I understand the advantage for a smaller api, but this will lead to different return types based on arguments, and complicated conditional logic inside the code.  It will make the code a lot harder to read, understand, and test, and also the api methods tougher to document and understand from the users perspective.  I think smaller more focused, testable methods with clear return types is better.

oveddan:
One way to help people pull what they want from the API, would be to extract some of the functionality contained within these methods into functions that are exported.  They don't necessarily need to be documented, but it would be useful for people to have access to.

For example, in my use case I just want the full image segmentation (not the pose).  I wrote code to accomplish this that looked like:

```typescript
import {BodyPix} from '@tensorflow-models/body-pix';
import * as posenet from '@tensorflow-models/posenet';
import * as tf from '@tensorflow/tfjs-core';
import {padAndResizeTo, scaleAndCropToInputTensorShape} from '@tensorflow-models/body-pix/dist/util';

export const estimateFullImageWithPersonSegmentations =
    (bodypixNet: BodyPix, input: tf.Tensor3D): tf.Tensor2D => {
      return tf.tidy(() => {
        const {resized, padding} =
            padAndResizeTo(input, bodypixNet.inputResolution);

        const {segmentation} = bodypixNet.baseModel.predict(resized);
        const [resizedHeight, resizedWidth] = resized.shape;
        const [height, width] = input.shape;

        const scaledSegmentation = scaleAndCropToInputTensorShape(
            segmentation, [height, width], [resizedHeight, resizedWidth],
            [[padding.top, padding.bottom], [padding.left, padding.right]],
            true);

        return scaledSegmentation.squeeze()
                   .clipByValue(0, 1)
                   .mul(tf.scalar(255, 'float32'))
                   .toInt() as tf.Tensor2D;
      });
    };
```

It would be nice if these pieces of functionality were encapsulated into functions that people could call if they want a specific piece of inference.

nsthorat:
@oveddan we can move that to another thread, for this PR I've renamed to the methods you suggested above:

segmentPerson segmentMultiPerson
segmentPersonParts segmentMultiPersonParts

PR #307, opened 2019-10-11
set default stride for MobileNet backbone to 16

tylerzhu-github:
This should fix the following bug:
https://github.com/tensorflow/tfjs/issues/2182

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/307)


tylerzhu-github:
Thanks Daniel. I am merging the fix now.

PR #311, opened 2019-10-13
[BodyPix] Import posenet types and decoding code into body-pix.

oveddan:
Fixes [#2183](https://github.com/tensorflow/tfjs/issues/2183)

The pose decoding code that body-pix v2 uses is the same as PoseNet's.  So instead of duplicating the code between the two folders, this PR adds PoseNet as a dependency and imports all of the decoding code and relevant types.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/311)


oveddan:
@nsthorat or @dsmilkov could you please publish to npm the latest posenet?  It has some functions that are exported that are required for this PR to work.

nsthorat:
Basically instead of

```
import {decodeMultiplePoses} from '@tensorflow-models/posenet';
// body pix src code
decodeMultiplePoses();
```

you would do

```
import {util} from '@tensorflow-models/posenet';
// body pix src code
util.decodeMultiplePoses();
```

This way we don't pollute the top-level namespace of posenet with things that are mostly meant to be internal (e.g. for vscode autocompletion of posenet).

oveddan:
Makes sense, sounds good.

PR #312, opened 2019-10-13
[BodyPix] - adds cpu decoding ops for person instance level decoding, and missing memory leak tests

oveddan:
* For instance level mask decoding, Checks the backend, and when the backend is not webgl, uses cpu decoding ops that were removed but retreived from old commits.
* adds missing dimensionality specification to decoding ops.
* renames files and functions to more clearly specify what they are doing.
* gets rid of repeated destructuring of padding in mask decoding methods to make code clearer.
* added tests that check for memory leaks for instance level decoding ops

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/312)

Fixes [#2187](https://github.com/tensorflow/tfjs/issues/2187) [#2189](https://github.com/tensorflow/tfjs/issues/2189)


oveddan:
This is a WIP because it appears as if padding of the resulting mask is off when using the CPU backend.

See the below images, which results when you set backend=cpu in the demo.

partMapPaddingOff
segmentationPaddingOff

@tylerzhu-github can you please look at the code and help debug this padding issue?  I just imported the code for cpu decoding from old commits, so don't fully understand what the code is doing.

oveddan:
Ok so, I wrote some code in node.js to test these methods.  The code runs bodypix instance decoding methods in node.js (cpu), then masks the image with the results.

Here is the source image:

tie_with_beer

Results of flattening the instance level person masks and masking the image:

person_mask

And results of flattening the instance level part masks for the left face part id, and applying that mask to the image:

part_mask


For reference, here is the code I wrote to test this:

```javascript
const bodypix = require('@tensorflow-models/body-pix')
const tf = require('@tensorflow/tfjs-node')
const fs = require('fs');
const {PNG} = require('pngjs');


const readFile = (fileName) => {
  return new Promise((resolve, reject) => {
    fs.readFile(fileName, (err, data) => {
      if (err) {
        reject(err);
      } else {
        resolve(data);
      }
    });
  });
};

const saveImage =
    async (file, image) => {
  const [height, width, depth] = image.shape;
  const asData = Array.from(await image.data());

  const pngImage = new PNG({width, height});

  pngImage.data = Buffer.from(asData);

  return new Promise((resolve) => {
    pngImage.pack().pipe(fs.createWriteStream(file)).on('close', () => resolve());
  });
};

async function getUint8ArrayFromImage(path) {
  const buf = await readFile(path);
  return new Uint8Array(buf);
}

const loadImageTensor = async(fileName) => {
  const uint8Array = await getUint8ArrayFromImage(fileName);

  return tf.node.decodeJpeg(uint8Array, 3);
};

function toFlatPersonMask(personSegmentations) {
  return tf.tidy(() => {
    const personSegmentationTensors = personSegmentations.map(x => tf.tensor2d(x.data, [x.height, x.width]));
    const tensorsAsMasks = personSegmentationTensors.map(x => x.greater(0.5).expandDims(0));

    const asOneTensor = tf.concat(tensorsAsMasks, 0);
    const asSingleMask = asOneTensor.sum(0).mul(tf.scalar(255, 'int32'));

    return asSingleMask;
  })
}

function toFlatPartSegmentationMaskForPart(personSegmentations, partId) {
  return tf.tidy(() => {
    const partSegmentationTensors = personSegmentations.map(x => tf.tensor2d(x.data, [x.height, x.width]));
    const tensorsAsMasks = partSegmentationTensors.map(x => x.equal(partId).expandDims(0));

    const asOneTensor = tf.concat(tensorsAsMasks, 0);
    const asSingleMask = asOneTensor.sum(0).mul(tf.scalar(255, 'int32'));

    return asSingleMask;
  });
}

async function main() {
  const image = await loadImageTensor('./tie_with_beer.jpg');

  console.log('loading bodypix');

  const bodyPixNet = await bodypix.load();

  console.log('estimating segmentation');

  const personSegmentations = await bodyPixNet.segmentMultiPerson(image);

  const personSegmentationMask = toFlatPersonMask(personSegmentations);

  const imageMasked = tf.concat3d([image, personSegmentationMask.expandDims(2)], 2);

  await saveImage('./person_mask.png', imageMasked);

  const partSegmentations = await bodyPixNet.segmentMultiPersonParts(image);

  const partSegmentationMask = toFlatPartSegmentationMaskForPart(partSegmentations, 0);

  const imageMaskedForPart = tf.concat3d([image, partSegmentationMask.expandDims(2)], 2);

  await saveImage('./part_mask.png', imageMaskedForPart);
}

main();
```

So I believe this is some bug in tf.js web cpu backend when it comes to cropping/paddings, as even the non instance level decoding methods result in off images.

nsthorat:
Hey Dan thanks for the awesome PR.

I think I am missing some context: what exactly do the two new methods for decoding masks do? Decode to what?

Is the motivation of this PR to get this to work in Node.js?

oveddan:
@nsthorat 

Yes, the main motivation is to get this to work in Node.js

For some context:
BodyPix 2.0 did decoding of the instance level masks using the CPU.  @tylerzhu-github did some awesome work converting this logic to webgl ops.  However, as a result of this, there is no corresponding op that runs on the CPU.  This PR restores the CPU decoding ops and uses those if the backend is not `webgl`.

The two new methods are the same as the existing methods, they just work on the CPU.

[This is the commit](https://github.com/tylerzhu-github/tfjs-models/commit/a01e24603f4976969ab2dc246c5eb2cc01e640cc) I pulled the CPU decoding code from.

Regarding adding the backend=cpu flag to the query string, I've been thinking about it and it's not ideal as it runs unbearably slow when this flag is enabled.

A better long-term solution would be to have a node.js demo that loads an image from the filesystem, performs inference on that image, and masks the image, saving the result to the filesystem.

oveddan:
Also, decoding instance level masks means that this is decoding a segmentation mask for each person that is detected.  Same goes for the body parts - this decodes a mask for each person for the body parts that are detected for that person.

nsthorat:
CPU stuff makes sense and I agree it should not be in the browser demo.

What is it decoding into? If I look at the README and public API today it seems like I have everything I need to do segmentation on a canvas so I must be missing something :)

oveddan:
The ops are part of the methods `segmentMultiPerson` or `segmentMultiPersonParts` - you cannot call those public api methods from node.js right now because they only work with a `webgl` backend.

Regarding *What is it decoding into?*:

Those two methods are taking a single segmentation mask for all people, and `decoding` them into a segmentation masks for each person, using the decoded poses and a long range offsets output to break the single segmentation mask into "instance level" ones.
`decode` is probably not the best method name starter.

nsthorat:
Got it -- but isn't that exactly what segmentMultiPerson does for you? Why would we need this to be exported?

oveddan:
Ah, so your main q is why does it need to be exported?   I guess it wouldn't need to be for most people.  

It's useful for people who would want to keep the tensors on the gpu and don't need the results downloaded to the CPU before being returned to the api.  i.e. if you pass a tensor to segmentMultiPerson, and you want a tensor back, there's no way to do it without uploading it back to the gpu.  If it's exported, *power* users could call this function with decoded outputs directly.

oveddan:
If it were to be used, then yes it would be used in concert with `segmentMultiPersonActivation`

My original motivation to exposed the `activation` functions were that using tensor ops to mask /blur images is more reliable than canvas compositing.  So I had some code that used the activation output to apply the mask, and this worked well in every browser, whereas canvas compositing did not work in many cases on safari ios.

That being said, at the moment I don't have a use for `segmentMultiPersonActivation` (I'm using the activation output of `segmentPerson` in my project, but thought in the future it would be nice to do some mobile friendly 'ar' like demos using the multi person output.  For those to work you would need the gpu based tensor.

I think there's a longer discussion in encapsulating some of the internals of these model inference methods into easier to consume parts, so that people that want to use them the way the api doesn't provide would find a way to do it.

tylerzhu-github:
Thx Dan! Really awsome PR and just a few comments.
Feel free to merge and we should be able to get this in before release 2.0

nsthorat:
Okay cool. Reviewing now.

In the future it's much easier to review if you make smaller PRs for a single feature or bug fix so we don't conflate these conversations. Maybe let's keep exporting those methods for a separate PR and just keep this PR for getting CPU decoding in Node.js working.

PR #296, opened 2019-09-23
Posenet - allow width and height specification for input resolution & relax input resolution requirement

oveddan:
This PR modifies the `posenet.load` function to allow on object containing `width` and `height` to be specified as input resolution, instead of just a single value.  Previously, specifying a single value always resulted in the image being resized to a square, causing a loss in accuracy.  See [#1844](https://github.com/tensorflow/tfjs/issues/1844) and [ml5 issue 510](https://github.com/ml5js/ml5-library/issues/510).

This also includes the changes from #293, which relaxes the requirement for the input resolution to be a specific value, and allows any value to be used;  internally the input resolution is adjusted to be a valid one, which is one that has its corresponding odd value divisible by the output stride.

Relaxing of the input resolution required values makes the API easier to consume and easier to understand. 

Additionally:
* Some "invalid" input resolutions were added to the demo as a form of test; internally those invalid resolutions will be adjusted to be valid resolutions.
* Added some tests around validation of the input resolution and setting it when the model is loaded.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/296)

PR #313, opened 2019-10-13
[BodyPix ]- allow input width and height to be specified for resolution.  Auto-set valid input resolution.

oveddan:
Relaxes input resolution validation to allow any number, and automatically sets the input resolution to something that is valid for the output stride.

Adds tests around input resolution.

Fixes [#2185](https://github.com/tensorflow/tfjs/issues/2185)
Inspired by #296

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/313)

### DIFF
--- a/body-pix/README.md
+++ b/body-pix/README.md
@@ -113,7 +113,7 @@ If you want to load other versions of the model, specify the architecture explic
 const net = await bodyPix.load({
   architecture: 'ResNet50',
   outputStride: 32,
-  inputResolution: 257,
+  inputResolution: { width: 640, height: 480 },
   quantBytes: 2
 });
 ```
@@ -123,7 +123,7 @@ const net = await bodyPix.load({
 const net = await bodyPix.load({
   architecture: 'MobileNetV1',
   outputStride: 16,
-  inputResolution: 513,
+  inputResolution: { width: 257, height: 200 },
   multiplier: 0.75
 });
 ```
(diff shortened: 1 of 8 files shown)


nsthorat:
qq -- why is inputResolution an argument to load? shouldn't it be part of inference?

nsthorat:
and shouldn't it also just be inferred based on the input size without an explicit argument?

oveddan:
I agree it should be an argument at inference time.

We had this discussion with Posenet 2.0 - where input resolution is not needed at load time, but just at inference time.  We decided not to change it to be at inference time in the end because we already released the API with it defined at load time, and didn't want to break it. Also there wouldn't be much of a use case for someone to change it at run time.

However, IMO it belongs in inference time because thats when it's used.  Also, changing the input resolution requires another call to load the model (which is visible in the demo).

Regarding adjusting the input resolution - the reason the user specifies this as an argument is that the image is resized to this resolution before inference.  It's not the actual input resolution of the image, its the input resolution that the image is scaled to before inference.

And one more thing - the input resolution is required to be adjusted to be divisible by the output stride for the pose decoding, which is only really needed for the `multi` methods, which do instance level segmentation mask decoding.

If we remove single pose decoding from the **non**-multi methods, [as suggested in this issue,](https://github.com/tensorflow/tfjs/issues/2196), then we just need to adjust the resolution for the `multi` methods.

nsthorat:
How hard would it be to move that logic to prediction time and remove it as a field? I think we should make that decision for you based on what you passed us (if you want a faster prediction time the user should make the image smaller before passing it to the model).

Regarding posenet API, that's fine, we can make the correct decisions here and then migrate them over to posenet and release another major version to break API.

oveddan:
The issue with requiring the user to scale the image first is that they need to scale the outputs back to the original input image size, which is non trivial.

They would need to scale up the segmentation masks, which we have the liberty of doing on the gpu using ops, but they wouldn't as it's already converted to be on the cpu.

They would also need scale up the poses for the `multi` methods.

oveddan:
Regarding moving to prediction time, it's not that difficult.

nsthorat:
Isn't that something we could do for them? e.g.

- Image comes in at [250, 250], we resize to [257, 257]
- Output is [257, 257] and we resize back to [250, 250]

oveddan:
@nsthorat yes that way is certainly doable.  

What happens though if the user has a 640x480 image, then they rescale it to 250x250 and then feed it as an input, and want the output back in the resolution of 640x480.

In the case of a mask output, they would need to scale that to the original resolution too, but without the luxury of having the tensor still on the gpu.
In the case of poses, they would have to call some code that scales it to the original resolution.

With posenet, we decided to add `imageScaleFactor` as a parameter because it made the api a lot easier to consume; we took care of scaling the image down, and scaling the outputs back to the original image size.  The sample code in the readme, and the demo code was a lot easier to read and duplicate as a result of encapsulating that feature within the api.

I can see the benefit for making the user do the resizing, as maybe its clearer for them what is actually happening.  The downside is they have more work to do as far as resizing the inputs and outputs.

nsthorat:
Okay that totally makes sense -- I like the scale factor approach of posenet better, I think it's clearer (and we can make it optional).

Can we do that here too (and this is during inference rather than model loading)?

oveddan:
Yeah image scale factor is great.  I'm for switching back to that!  Sounds good.  
Should it default to 0.5 like how it was before?

oveddan:
@nsthorat so what I will do is open another PR with `imageScaleFactor` as an argument at inference time, instead of taking inputResolution at load time.

I'm out of town for about a week, I think in the meantime, since this PR just improves the existing functionality and doesnt break the API, as well as adds some tests, if it looks good to you it would make sense to merge it.  Then I can do another PR that changes to imageScaleFactor at runtime.

What do you think?

oveddan:
Closing as a result of #321 
Will open another PR that takes care of maintaining the aspect ratio

PR #321, opened 2019-10-17
Refactor `inputResolution` to `internalResolution`

tylerzhu-github:
This is the final PR before releasing BodyPix 2.0

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/321)


nsthorat:
Retrying CI for this, looks like speech commands tests are flaky @caisq: https://pantheon.corp.google.com/cloud-build/builds/1e277170-4770-42c5-894a-25ddc0948a7e;step=1?project=learnjs-174218

oveddan:
This is a great improvement.  I really believe though this should be a percentage underneath and not a hardcoded resolution.  What if the user wants to do a full 1080p image?

oveddan:
This will also result in the image being resized to a square - causing black pixels to be added and resulting in a loss in accuracy.

oveddan:
@nsthorat @tylerzhu-github what do you think of doing another PR that uses a percentage instead of a hardcoded resolution value?  This will allow the original image to maintain its aspect ratio without adding unnecessary black padding.

PR #325, opened 2019-10-21
Fix security warnings.

nsthorat:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/325)


nsthorat:
Yep, actually the posenet demo doesn't work because of one of the inputResolution changes and knn classifier has a bug at head but parcel bundler doesn't change anything (tfjs still runs).

PR #330, opened 2019-10-22
Dynamically determines input height and width

tylerzhu-github:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/330)


body-pix/src/output_rendering_util.ts
```diff
@@ -348,16 +348,16 @@ const CANVAS_NAMES = {
...
+  canvas.width = image.width;
+  canvas.height = image.height;
 
   const ctx = canvas.getContext('2d');
   ctx.save();
   if (flipHorizontal) {
     flipCanvasHorizontal(canvas);
   }
 
-  ctx.drawImage(image, 0, 0, image.offsetWidth, image.offsetHeight);
```
nsthorat: use .width here dont remove these

body-pix/src/output_rendering_util.ts
```diff
@@ -399,8 +399,7 @@ export function drawPixelatedMask(
     maskOpacity = 0.7, maskBlurAmount = 0, flipHorizontal = false,
     pixelCellWidth = 10.0) {
   assertSameDimensions(
-      {width: image.offsetWidth, height: image.offsetHeight}, maskImage,
```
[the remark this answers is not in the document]
tylerzhu-github: done

body-pix/src/output_rendering_util.ts
tylerzhu-github: done

oveddan:
___
*[body-pix/src/util.ts, line 8 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/330#-LrpAWsf1bdPEIWCY993:-LrpAWsf1bdPEIWCY994:b-c57j7z) ([raw file](https://github.com/tensorflow/tfjs-models/blob/a2e47ff34bc79ece119f7fd4ebb5c5317b181c75/body-pix/src/util.ts#L8)):*
> ```XML
> 
> export function getInputSize(image: BodyPixInput) {
>   if (image instanceof HTMLImageElement || image instanceof HTMLCanvasElement) {
> ```

What if the image is hidden? (display:none) ?

oveddan:
___
*[body-pix/src/util.ts, line 15 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/330#-LrpAf26CiKcz8m_E_fj:-LrpAf26CiKcz8m_E_fk:b-cx5wk2) ([raw file](https://github.com/tensorflow/tfjs-models/blob/a2e47ff34bc79ece119f7fd4ebb5c5317b181c75/body-pix/src/util.ts#L15)):*
> ```XML
>     return [image.videoHeight, image.videoWidth];
>   } else {
>     return [image.shape[0], image.shape[1]]
> ```

I would check that it is an instance of a tf.Tensor, and if not, then throw an error at the end

oveddan:
___
*[body-pix/src/body_pix_model.ts, line 436 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/330#-LrpABx7CvDQCzzwvThH:-LrpABx7CvDQCzzwvThI:bp8vayy) ([raw file](https://github.com/tensorflow/tfjs-models/blob/a2e47ff34bc79ece119f7fd4ebb5c5317b181c75/body-pix/src/body_pix_model.ts#L436)):*
> ```XML
>     padding: Padding
>   } {
>     const [height, width] = getInputSize(input);
> ```

Do we no longer need to convert the input to a tensor?  If so, that's great

oveddan:
___
*[body-pix/src/util.ts, line 13 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/330#-LrpBER92A74SGk3Vu_l:-LrpB3VD5jbr_BuYIkmh:b-qm7bcy) ([raw file](https://github.com/tensorflow/tfjs-models/blob/a2e47ff34bc79ece119f7fd4ebb5c5317b181c75/body-pix/src/util.ts#L13)):*
> ```XML
>     return [image.height, image.width];
>   } else if (image instanceof HTMLVideoElement) {
>     return [image.videoHeight, image.videoWidth];
> ```

This here is great!  No longer need to set the width and height explicitly.

The only thing is - if the user sets the width and height on the element, it will be overriden by the videoWidth and videoHeight.  So maybe first check if those are set, and if not, use the videoWidth and videoHeight?

I don't feel strongly about this, just curious as to what we should do in that case

tylerzhu-github:
The PR is ready to be merged.

PR #332, opened 2019-10-22
Run tests on browserstack for body pix.

nsthorat:
This does not change any existing tests just runs the existing ones on browserstack instead of in node.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/332)

### DIFF
--- a/body-pix/cloudbuild.yml
+++ b/body-pix/cloudbuild.yml
@@ -31,6 +31,10 @@ steps:
   waitFor: ['yarn']
 
 # General configuration
+secrets:
+- kmsKeyName: projects/learnjs-174218/locations/global/keyRings/tfjs/cryptoKeys/enc
+  secretEnv:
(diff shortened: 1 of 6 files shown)


oveddan:
___
*[body-pix/src/setup_test.ts, line 42 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/332#-Lroz8e_3ML9Zk6CO0Yb:-LrozzAHBUfbqn5uwWNe:b-e98zk7) ([raw file](https://github.com/tensorflow/tfjs-models/blob/b60da341271b15ca673de59a93690870d7b8fafc/body-pix/src/setup_test.ts#L42)):*

done
</blockquote></details>

As master stands now - tests still pass on the cpu, we just don't have tests those methods.
So uncommenting this would still allow the tests to pass on the cpu.

Once this PR is reviewed and merged:
https://github.com/tensorflow/tfjs-models/pull/312

It adds missing tests for the api methods that don't have cpu based ops.

PR #334, opened 2019-10-25
Mentions pose in the README-v2.md of BodyPix

tylerzhu-github:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/334)


tylerzhu-github:
> Since we said that the non instance level methods should return multi pose, but given that multi-pose decoding has not been implemented yet for those methods, how about having those api methods return an array of poses but just contain the single pose, and in the future we can do another PR that switches it to decode multiple poses?
> 
> This way we don't break the api when this is changed.
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/334#-:-LsSeJ0h9G-RvJd0LGjs:b-eq3s7f)_ status: 0 of 1 approvals obtained

This PR contains README changes. We can add an addition field "allPoses" to the returned "dictionary" and make sure its backward compatible in the next PR when addressing the issue you mentioned. The "pose" field will still be used by the "multiPerson" methods.

PR #336, opened 2019-10-26
[speech-commands] Fix memory leak in the recognize() method

caisq:
BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/336)


caisq:
cc @irealva

dsmilkov:
Nice fix!!

PR #349, opened 2019-11-08
Fix datasets urls in deeplab README

phc:
1. Fixes the relative urls (path starts from the current directory instead of root)
2. Changes the lines number to jump to labels instead of color maps

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/349)


nsthorat:
Can you accept the CLA then I will merge? Thanks!

phc:
@googlebot I signed it!

phc:
@nsthorat done! Thanks

PR #357, opened 2019-11-15
[BodyPix] - increase max internal resolution to 2

oveddan:
For BodyPix - increase max internal resolution to 2, so that inference can be run to emulate a lower output stride and get higher accuracy.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/357)

### DIFF
--- a/body-pix/src/util.ts
+++ b/body-pix/src/util.ts
@@ -58,7 +58,7 @@ const INTERNAL_RESOLUTION_PERCENTAGES = {
   [INTERNAL_RESOLUTION_STRING_OPTIONS.full]: 1.0
 };
 
-const MAX_INTERNAL_RESOLUTION = 4;
+const MAX_INTERNAL_RESOLUTION = 2;
 
 function toInternalResolutionPercentage(
     internalResolution: BodyPixInternalResolution): number {
--- a/body-pix/src/util.ts
+++ b/body-pix/src/util.ts
@@ -58,6 +58,8 @@ const INTERNAL_RESOLUTION_PERCENTAGES = {
   [INTERNAL_RESOLUTION_STRING_OPTIONS.full]: 1.0
 };
 
+const MAX_INTERNAL_RESOLUTION = 4;
+
 function toInternalResolutionPercentage(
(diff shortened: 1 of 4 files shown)


body-pix/src/util.ts
```diff
@@ -71,10 +74,11 @@ function toInternalResolutionPercentage(
...
   } else {
     tf.util.assert(
-        typeof internalResolution === 'number' && internalResolution < 1 &&
-            internalResolution > 0,
+        typeof internalResolution === 'number' &&
+            internalResolution <= MAX_INTERNAL_RESOLUTION &&
+            internalResolution >= MIN_INTERNAL_RESOLUTION,
         () =>
-            `inputResolution must be a string or number between 0 and 1, but ` +
+            `inputResolution must be a string or number between 0 and 4, but ` +
```
nsthorat: this doesn't match the constant -- and let's use the constant in the error instead of hardcoding the string.

oveddan:
___
*[body-pix/src/util.ts, line 81 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/357#-LtjXl4pNASQXXcpPg2T-r2-81:-LvCMzXxCd_87cFwD0mi:bz8nf0i) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9da13d5d88c7a76fe1809513c519399d916f869e/body-pix/src/util.ts#L81)):*

this doesn't match the constant -- and let's use the constant in the error instead of hardcoding the string.
</blockquote></details>

Good call.  Fixed

oveddan:
___
*[body-pix/src/util.ts, line 81 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/357#-LtjXl4pNASQXXcpPg2T-r2-81:-LvCNIa17fhe5ogCWYj4:bdmf86b) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9da13d5d88c7a76fe1809513c519399d916f869e/body-pix/src/util.ts#L81)):*

Good call.  Fixed
</blockquote></details>

Fixed in PR #365

PR #362, opened 2019-11-29
 fix default main path of package.json 

willin:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/362)


willin:
@googlebot I signed it!

jenperson:
`dist/index.js` is the primary entry point for the package.

PR #381, opened 2019-12-19
Actually use TFHub for USE and toxicity models.

annxingyuan:
The lockfiles for the demos need to be updated after the modules are actually published.

I thought the workflow could be: once I get approval on this PR, I'll publish both USE and toxicity to NPM. Then I'll upgrade the yarn.lock files for both demos. Then I'll merge into master.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/381)

### DIFF
--- a/toxicity/src/index.ts
+++ b/toxicity/src/index.ts
@@ -20,9 +20,6 @@ import * as tfconv from '@tensorflow/tfjs-converter';
 import * as tf from '@tensorflow/tfjs-core';
 export {version} from './version';
 
-const BASE_PATH =
-    'https://storage.googleapis.com/tfjs-models/savedmodel/toxicity/';
-
 /**
  * Load the toxicity model.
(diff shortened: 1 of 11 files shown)


toxicity/src/index.ts
```diff
@@ -52,7 +49,9 @@ export class ToxicityClassifier {
   }
 
   async loadModel() {
-    return tfconv.loadGraphModel(`${BASE_PATH}model.json`);
+    return tfconv.loadGraphModel(
+        'https://tfhub.dev/tensorflow/tfjs-model/toxicity/1/default/1',
```
sorensenjs: Are you loading the original model without training for the particular task?

annxingyuan: It's the same model we've been serving, except now the assets live on TFHub.

PR #392, opened 2020-01-16
[blazeface] Add wasm, add flipHorizontal flag, expose more types.

annxingyuan:
Updated demo: https://storage.googleapis.com/tfjs-models/demos/blazeface/index.html

After this is merged / publish I'll update the demo yarn.lock file.

#### Changes
- Default backend to `wasm` in demo and add GUI for switching backends.
- Add `flipHorizontal` flag.
- Expose `BlazeFaceModel` type.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/392)


annxingyuan:
Hey @dsmilkov - thanks for the LGTM but could you take another look? I left a few comments and just want to make sure we're on the same page.

PR #397, opened 2020-01-30
[blazeface] Enable getting bounding boxes only, export more types.

annxingyuan:
Facemesh will depend on blazeface for bounding box detection. This PR makes it possible to fetch only the bounding boxes from blazeface without paying for other properties like facial landmarks.

Will update demo yarn.lock file after publishing blazeface, then will merge.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/397)


annxingyuan:
@nsthorat @dsmilkov sending a ping on this PR :)

PR #414, opened 2020-03-02
[handpose] Add model.

annxingyuan:
#### Notes
- Unlike for facemesh, almost all the post-processing for handpose happens on the CPU because the tensors are much smaller, and I found that converting things to the CPU improved performance. (I also tested the treatment for facemesh, in that case moving postprocessing to the CPU slows things down.)
- There's no `returnTensors` API because we're working with JS arrays in the first place.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/414)


niltype22:
Bumping multi-hand support

PR #423, opened 2020-03-21
Update bodypix version to 2.0.5.

nsthorat:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/423)


dsmilkov:
do you want to release and update the demo/package.json in the same PR or separate? up to you

nsthorat:
I was going to avoid doing that since this is just a types patch fix.

PR #427, opened 2020-03-24
[handpose] Add rotate op

axinging:
Rotate op is required in handpose when running in webgpu backend.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/427)

### DIFF
--- a/handpose/demo/index.js
+++ b/handpose/demo/index.js
@@ -18,7 +18,14 @@
 import * as tfwebgpu from '@tensorflow/tfjs-backend-webgpu';
 import * as tf from '@tensorflow/tfjs-core';
 import * as handpose from '@tensorflow-models/handpose';
+import * as tfjsWasm from '@tensorflow/tfjs-backend-wasm';
+// TODO(annxingyuan): read version from tfjsWasm directly once
+// https://github.com/tensorflow/tfjs/pull/2819 is merged.
+import {version} from '@tensorflow/tfjs-backend-wasm/dist/version';
 
+tfjsWasm.setWasmPath(
+    `https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm@${
+        version}/dist/tfjs-backend-wasm.wasm`);
 function isMobile() {
   const isAndroid = /Android/i.test(navigator.userAgent);
   const isiOS = /iPhone|iPad|iPod/i.test(navigator.userAgent);
@@ -41,7 +48,9 @@ const mobile = isMobile();
 // to avoid crowding limited screen space.
 const renderPointcloud = mobile === false;
 
-const state = {};
+const state = {
+  backend: 'webgl'
+};
 
 if (renderPointcloud) {
   state.renderPointcloud = true;
@@ -58,6 +67,20 @@ function setupDatGui() {
   }
 }
 
+function setupDatGui() {
+  const gui = new dat.GUI();
+  gui.add(state, 'backend', ['wasm', 'webgl', 'cpu', 'webgpu'])
+      .onChange(async backend => {
+        await tf.setBackend(backend);
+      });
+
+  if (renderPointcloud) {
+    gui.add(state, 'renderPointcloud').onChange(render => {
+      document.querySelector('#scatter-gl-container').style.display =
+          render ? 'inline-block' : 'none';
+    });
+  }
+}
 function drawPoint(ctx, y, x, r) {
   ctx.beginPath();
   ctx.arc(x, y, r, 0, 2 * Math.PI);
@@ -130,7 +153,7 @@ async function loadVideo() {
 }
 
 const main = async () => {
-  await tf.ready();
+  await tf.setBackend(state.backend);
   model = await handpose.load();
   let video;
 
--- a/handpose/demo/index.js
+++ b/handpose/demo/index.js
@@ -15,6 +15,8 @@
  * =============================================================================
  */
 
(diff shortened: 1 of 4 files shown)


axinging:
@annxingyuan PTAL

axinging:
> @axinging this is awesome :) could you also add code in handpose/src/pipeline.ts that actually invokes this kernel if the user is running the webgpu backend?

Indeed I tried to apply below code before(which can be used to invoke this kernel). But this may fail on webgl.
```
--- a/handpose/src/pipeline.ts
+++ b/handpose/src/pipeline.ts
+import {rotate as rotateWebgpu} from './rotate_webgpu';
@@ -187,6 +188,8 @@ export class HandPipeline {
     if (backend === 'webgl') {
       rotatedImage = rotateWebgl(image, angle, 0, palmCenterNormalized);
+    } else if (backend === 'webgpu') {
+      rotatedImage = rotateWebgpu(image, angle, 0, palmCenterNormalized);
```
On WebGPU (with await tf.ready(); applied), this works well.
While on WebGL, this complains:
```
backend_webgpu.ts:96 Uncaught ReferenceError: GPUBufferUsage is not defined
    at Object.parcelRequire.node_modules/@tensorflow-models/handpose/node_modules/@tensorflow/tfjs-backend-webgpu/dist/backend_webgpu.js../flags_webgpu (backend_webgpu.ts:96)
    at newRequire (demo.e31bb0bc.js:47)
    at localRequire (demo.e31bb0bc.js:53)
    at Object.parcelRequire.node_modules/@tensorflow-models/handpose/node_modules/@tensorflow/tfjs-backend-webgpu/dist/index.js../flags_webgpu (index.ts:23)
    at newRequire (demo.e31bb0bc.js:47)
    at localRequire (demo.e31bb0bc.js:53)
    at Object.parcelRequire.node_modules/@tensorflow-models/handpose/dist/handpose.esm.js.@tensorflow/tfjs-core (handpose.esm.js:17)
    at newRequire (demo.e31bb0bc.js:47)
    at localRequire (demo.e31bb0bc.js:53)
    at Object.parcelRequire.index.js.@tensorflow-models/handpose (index.js:18)

```

annxingyuan:
@axinging Sorry I don't quite understand - are you saying we need to add `await tf.ready()` to the model initialization code in order to make it work with WebGPU? Would you mind adding that to the handpose model? What do you mean that the code fails on WebGL? This codepath should not be hit if the user is using the WebGL backend, no?

axinging:
@annxingyuan 
Yes，agree that the code path of rotateWebgpu should not be hit in runtime, in WebGL.

I guess the problems is caused by the import chain:
In handpose/src/pipeline.ts, it imports rotate_webgpu.ts
```
import {rotate as rotateWebgpu} from './rotate_webgpu';
```
In rotate_webgpu.ts, it imports tfjs-backend-webgpu:
```
import * as tfwebgpu from '@tensorflow/tfjs-backend-webgpu';
```
I noticed that there is a method to "[autoremove-all-unused-imports](https://wesleygrimes.com/angular/2019/02/14/how-to-use-tslint-to-autoremove-all-unused-imports-in-a-typescript-project.html)",  but this method is used at compile time, not runtime.

As to my issue, the backend is decided at runtime. So this import dependences can not be removed at compile time. That's why localRequire of webgpu happens in WebGL.

And, I think we need work out a method to test the operator add by tfjs models.

annxingyuan:
@axinging oh I see - so we need to add a check in the WebGPU backend initialization code to make sure we don't error out if the browser doesn't support WebGPU - right? Would be great if you could file a bug for this! 

What do you mean by 'provide a flag, to let user select which backend'? Do you mean the handpose model should expose a flag that controls which backend is used? We usually do this outside the model, e.g. https://github.com/tensorflow/tfjs-models/blob/master/facemesh/demo/index.js#L172 - or did I misunderstand your suggestion?

axinging:
@annxingyuan  
Issue is hree :https://github.com/tensorflow/tfjs/issues/3003

Yes, the flage at https://github.com/tensorflow/tfjs-models/blob/master/facemesh/demo/index.js#L172 is what I suggest.

axinging:
@annxingyuan, hi I add the tf.ready, PTAL

axinging:
> Thank you @axinging ! I'm excited to try this out. Could you add the option to toggle between WebGL and WebGPU in the demo gui? You can see an example of how we do this in the facemesh demo (there we have the ability to toggle between WebGL and WASM).

@annxingyuan , PTAL, added the backend controls.
Noted, currently only CPU/WebGL backend works.

handpose/demo/index.js
```diff
@@ -15,8 +15,17 @@
...
  */
 
+import * as tfwebgpu from '@tensorflow/tfjs-backend-webgpu';
+import * as tf from '@tensorflow/tfjs-core';
 import * as handpose from '@tensorflow-models/handpose';
-
+import * as tfjsWasm from '@tensorflow/tfjs-backend-wasm';
+// TODO(annxingyuan): read version from tfjsWasm directly once
+// https://github.com/tensorflow/tfjs/pull/2819 is merged.
+import {version} from '@tensorflow/tfjs-backend-wasm/dist/version';
```
annxingyuan, line 24: #2819 should have been merged, so you should be able to read the version directly from tfjsWasm now instead of fetching it from dist.

axinging: Thanks Ann, I followed up it here: https://github.com/tensorflow/tfjs-models/pull/442

PR #429, opened 2020-03-26
Improve README's.

annxingyuan:
Fix typos and use vanilla syntax for importing modules.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/429)


annxingyuan:
@nsthorat @dsmilkov sandeep pointed out that the import statements don't work without something like parcel - LMK if you think this change is a good idea and I can change the import statements for our other models as well.

PR #435, opened 2020-04-15
1498 knn memory leak

forina20:
Awesome!

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/435)


jenperson:
It looks like this change has been made in the code, so the PR is being closed.

PR #436, opened 2020-04-15
Bump dev deps version to remove security vulnerability

kangyizhang:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/436)


wwy2018:
hi, i tried to use toxicity in a vue project, then met with the problem that the model couldn't get loaded. looking forward to know the right way to use the model in a html page.

PR #449, opened 2020-05-20
Add return type overloads for estimateFaces

kevmo314:
This improves the ergonomics when a constant true/false value is passed to the function. TypeScript will choose the correct return type accordingly.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/449)


kevmo314:
@googlebot I signed it!

jenperson:
Closing because the package has since been deprecated.

PR #464, opened 2020-07-04
[SpeechCommands] Add notebook to show how to convert model to tflite

caisq:
* Fixes https://github.com/googlecreativelab/teachablemachine-community/issues/36

This change is [](https://reviewable.io/reviews/tensorflow/tfjs-models/464)


caisq:
Still needs some work. Please hold off review for now. I'll assign reviewer(s) when this PR is ready for review.

caisq:
cc @lintian06

speech-commands/training/browser-fft/tflite_conversion.ipynb
```diff
@@ -0,0 +1,346 @@
+{
```
pyu10055, line 1: <p>please do not use the rc0 version, it is referencing the nightly estimator build.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/tfjs-models/pull/464/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

pyu10055, line 1: <p>Is this the older model format? will it be better to upgrade the model with the latest version of tfjs?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/tfjs-models/pull/464/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

caisq: <p>I agree we should avoid release candidates in general. However, in this case I have to use 2.3.0rc0. This is the early non-nightly version in which this tensorflow bug is fixed: <a href="https://github.com/tensorflow/tensorflow/issues/38135" target="_blank">https://github.com/tensorflow/tensorflow/issues/38135</a></p><p> </p><p>When the final release of 2.3.0 comes out, I'll switch to that version.</p>

caisq: <p>This is a layers model. So it's already the latest format because there is no newer version of the format AFAIK. </p><p> </p><p>However, the model json is indeed in an old format, in that the "config" field of the "sequential" model is an array instead of a dict as in the latest format. But both tensorflow and tensorflowjs ought to be backward compatible to this old format. Also, users may get this old format from Teachable Machine, so we should make sure we can handle it correctly.</p><p> </p>

caisq: Thanks for the review, @pyu10055 

PTAL.

pyu10055: please add a TODO for this, afaik the rc build is broken somewhat too.

caisq: Done.

PR #467, opened 2020-07-10
[facemesh] Ensure bounding box is returned correctly

SevInf:
The code assumes corners of the bounding box have type [number, number]
when the actual type is [[number, number]]. As a result, when
flipHorizontal is true, corners of a bounding box will always evaluate
to [NaN, undefined].

Fixed by ensuring corners actually typed as [number, number].

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/467)


annxingyuan (approved):
Thank you @SevInf ! LGTM

annxingyuan:
@SevInf ah I realized in the time it took for me to review your PR we actually moved the bounding box calculations to the CPU which fixed some of these issues. However the test is still valid so I will merge that in :)

PR #468, opened 2020-07-10
[facemesh] Correctly work if webgl backend is not registred

SevInf:
env.get(WEBGL_PACK_DEPTHWISECONV) throws an exception if webgl backend
is not registred. This makes facemesh impossible to use without shipping
this backend, even when the aplication intends to use other backends.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/468)


annxingyuan:
Hey @SevInf - thank you so much for this contribution! I'm sorry it has taken me so long to review it - in the meantime the underlying issue has actually been fixed: https://github.com/tensorflow/tfjs-models/blob/master/facemesh/src/index.ts#L219 - I wish I had seen your PR sooner :)

Thanks again.

PR #486, opened 2020-07-19
[facemesh] Support rotated faces, improve docs.

annxingyuan:
Currently, the facemesh model fails to detect faces that are rotated beyond ~10 degrees. This PR adds a rotation step to the post processing pipeline that improves the model's ability to detect rotated faces.

The new post processing pipeline involves many operations on small tensors. Therefore, I've moved most of the post processing logic to the CPU, similar to handpose.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/486)


annxingyuan:
@pyu10055 @lina128 Hey - sorry I realize you don't have much context on this PR :) Please let me know if I can explain anything in more detail! Basically this PR adds logic to rotate the detected faces before they get passed to the mesh detection model, because the mesh detection model was only trained on vertically oriented faces. 

There are a lot of code changes because implementing the rotation logic in WebGL is worse for performance, so we have to move a lot of data into the CPU as part of this PR.

dosomder:
After updating to 0.0.4 I don't have typings support anymore. Looks like the file index.d.ts is not present in 0.0.4, but when I downgrade to 0.0.3 it's there.

dosomder:
@annxingyuan @tafsiri Any reason typings were removed?

annxingyuan:
Hi @dosomder - sorry for the delay in getting back to you - the typings were inadvertently removed as a result of TypeScript configuration updates between 0.0.4 and 0.0.3. They will be restored in the next release: https://github.com/tensorflow/tfjs-models/pull/508

annxingyuan:
@dosomder for more context, this is the PR we just merged for reviving declarations: https://github.com/tensorflow/tfjs-models/pull/522

ayyar:
> @dosomder for more context, this is the PR we just merged for reviving declarations: #522

@annxingyuan  any idea when the typings fix will be pushed out?

annxingyuan:
@ayyar We should have a new facemesh release in about a week.

PR #488, opened 2020-07-24
Support passing model URL as  parameters

codedream4u:
Sometimes we need store the models in an inner-network and should customized the URL of models.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/488)

### DIFF
--- a/handpose/src/index.ts
+++ b/handpose/src/index.ts
@@ -23,29 +23,25 @@ import {MESH_ANNOTATIONS} from './keypoints';
 import {Coords3D, HandPipeline, Prediction} from './pipeline';
 
 // Load the bounding box detector model.
-async function loadHandDetectorModel() {
-  const HANDDETECT_MODEL_PATH =
-      'https://tfhub.dev/mediapipe/tfjs-model/handdetector/1/default/1';
-  return tfconv.loadGraphModel(HANDDETECT_MODEL_PATH, {fromTFHub: true});
+async function loadHandDetectorModel(handdetectModelURL:string) {
+  return tfconv.loadGraphModel(handdetectModelURL, {fromTFHub: true});
 }
 
 const MESH_MODEL_INPUT_WIDTH = 256;
 const MESH_MODEL_INPUT_HEIGHT = 256;
 
 // Load the mesh detector model.
-async function loadHandPoseModel() {
-  const HANDPOSE_MODEL_PATH =
-      'https://tfhub.dev/mediapipe/tfjs-model/handskeleton/1/default/1';
-  return tfconv.loadGraphModel(HANDPOSE_MODEL_PATH, {fromTFHub: true});
+async function loadHandPoseModel(handposeModelURL:string) {
+  return tfconv.loadGraphModel(handposeModelURL, {fromTFHub: true});
 }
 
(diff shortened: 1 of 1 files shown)


handpose/src/index.ts
```diff
@@ -67,15 +63,21 @@ interface AnnotatedPrediction extends Prediction {
...
  */
 export async function load({
   maxContinuousChecks = Infinity,
   detectionConfidence = 0.8,
   iouThreshold = 0.3,
-  scoreThreshold = 0.5
+  scoreThreshold = 0.5,
+  handdetectModelURL = 'https://tfhub.dev/mediapipe/tfjs-model/handdetector/1/default/1',
+  handposeModelURL = 'https://tfhub.dev/mediapipe/tfjs-model/handskeleton/1/default/1',
+  anchorsURL = 'https://tfhub.dev/mediapipe/tfjs-model/handskeleton/1/default/1/anchors.json?tfjs-format=file'
```
annxingyuan, line 77: can we call these `handdetectorModelURL` and `handskeletonModelURL`?

handpose/src/index.ts
```diff
@@ -23,29 +23,25 @@ import {MESH_ANNOTATIONS} from './keypoints';
 import {Coords3D, HandPipeline, Prediction} from './pipeline';
 
 // Load the bounding box detector model.
-async function loadHandDetectorModel() {
-  const HANDDETECT_MODEL_PATH =
-      'https://tfhub.dev/mediapipe/tfjs-model/handdetector/1/default/1';
-  return tfconv.loadGraphModel(HANDDETECT_MODEL_PATH, {fromTFHub: true});
+async function loadHandDetectorModel(handdetectModelURL:string) {
+  return tfconv.loadGraphModel(handdetectModelURL, {fromTFHub: true});
```
annxingyuan, line 27: If you pass in a custom URL, we probably shouldn't pass in {fromTFHub: true} here.

handpose/src/index.ts
stevenjunliu: I think URL is a broad concept and `path` is a type of URL.
In `loadGraphMode` function,  it is called `modelURL` and supports google storage URL or tfhub path.

handpose/src/index.ts
stevenjunliu: Thank you. 

From the `loadGraphModel` document:
const modelUrl = 'https://storage.googleapis.com/tfjs-models/savedmodel/mobilenet_v2_1.0_224/model.json';
const model = await tf.loadGraphModel(modelUrl);

const modelUrl =  'https://tfhub.dev/google/imagenet/mobilenet_v2_140_224/classification/2';
const model = await tf.loadGraphModel(modelUrl, {fromTFHub: true});

When passing `loadOptions` =  {fromTFHub: true}, it just means the modelURL ends with a folder ?

So I suppose the easiest way to clear away confusion is  refining the comments of `load` function?  like:
- `handdetectModelURL`  Specify  tfhub mirror site URL of handdetect model. Defaults to tfhub.dev.

By the way, other tfjs-models like `speech-commands`, the custom URL is similar to storage.google.

jenperson:
Thank you for your contribution! This PR is being closed because the requested changes where not made.

PR #500, opened 2020-08-04
revert default tsconfig settings to output commonjs+es5

tafsiri:
This PR restores build settings for tfjs-models to target es5 by default. This allows the existing rollup configs to work without modification even if we upgrade the dependencies to depend on 2.x (such as cocoSsd does [here](https://github.com/tensorflow/tfjs-models/pull/462)).

I don't think we need to upgrade models to compile their code to es2017. Unlike the main library we are not trying to make them tree-shakeable.

PTAL and let me know if there are configurations you don't think are well covered or if there is a way to produce any bugs that came up while doing [this upgrade](https://github.com/tensorflow/tfjs-models/pull/462). We can do pre-releases if we want to test in specific programs. You can also try it locally and link to the node script we were talking about earlier. 

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/500)


tafsiri:
Tested USE locally in node and via the demo page.

tafsiri:
If this approach overall looks good we should revert the rollup settings for cocoSsd to match. Or we could have custom settings for cocoSsd. Any preference?

tafsiri:
@pyu10055 I updated cocoSsd package.json and rollup to generate the esm file for module and jsnext:main sections.

tafsiri:
@pyu10055 merged. you should be good to go on releasing cocoSsd and use (from a bundle perspective).

PR #508, opened 2020-08-10
[facemesh] Add iris detection support. Fix rotation bug.

annxingyuan:
This also fixes https://github.com/tensorflow/tfjs/issues/3827

This PR adds support for detecting iris keypoints through the MediaPipe iris model. The iris model returns the positions of the iris as well as the contours around the eyes. We update the mesh returned by the facemesh model with the iris keypoints when the user is facing the camera directly. If the user is turned away from the camera the mesh tends to diverge too much from the iris keypoints, so we replace only the eyelid. 

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/508)


annxingyuan:
@wakoan Hi Andrey!

startupgurukul:
Currently iris seems to be run after landmark detection, since blazeface also returns eye coordinates it may be little faster if landmarks and iris are kicked off at a time

Also with iris, it takes very long for the model to predict  on a mobile device 

Do you think adding a flag, 
1).   with iris
2).  without iris 
3). only iris

startupgurukul:
One more request, landmarks+ iris, it takes very long for the model to predict  on a mobile device 

Do you think adding a flag, 
1).   with iris
2).  without iris 
3). only iris, do consider

pyu10055 (approved):
> Reviewed 6 of 13 files at r2, 1 of 3 files at r3, 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/508)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @lina128, @pyu10055, and @tafsiri)

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/508)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @annxingyuan, @lina128, @pyu10055, and @tafsiri)

___
*[facemesh/src/index.ts, line 85 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/508#-MGk6PM60LSfy1v9eIrI:-MGk6PM7-rJTUPDeU8In:bk7om6d) ([raw file](https://github.com/tensorflow/tfjs-models/blob/15619fb7ebe32d0b23afa66a0ee79978f1fff071/facemesh/src/index.ts#L85)):*
> ```XML
>   scoreThreshold = 0.75
> } = {}): Promise<FaceMesh> {
>   const [blazeFace, blazeMeshModel, irisModel] = await Promise.all([
> ```

How big is the iris model? should there be a flag for controlling if the irisModel should be loaded? Since user could be only using balzemesh model without iris detection.
___
*[facemesh/src/index.ts, line 217 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/508#-MGk6yQwEM7lsrT6GGgU:-MGk6yQwEM7lsrT6GGgV:belqway) ([raw file](https://github.com/tensorflow/tfjs-models/blob/15619fb7ebe32d0b23afa66a0ee79978f1fff071/facemesh/src/index.ts#L217)):*
> ```XML
>       HTMLCanvasElement,
>       returnTensors = false, flipHorizontal = false,
>       predictIrises = true): Promise<AnnotatedPrediction[]> {
> ```

should iris detection be default to false, since enabling it has a big performance hit?

annxingyuan (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/508)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128, @pyu10055, and @tafsiri)

___
*[facemesh/demo/index.js, line 63 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/508#-MFkXh0xAYWTnewybHAR:-MGsZFTmAUeCCJK6bE5E:b-r7x3bi) ([raw file](https://github.com/tensorflow/tfjs-models/blob/622478d29f8b2e2e0585834443898ceb367f2bb8/facemesh/demo/index.js#L63)):*
<details><summary><i>Previously, annxingyuan (Ann Yuan) wrote…</i></summary><blockquote>

Yes, we have to run inference on an additional model in order to obtain the iris coordinates. On WebGL, predicting irises means the model runs at ~15fps. Without iris prediction the model runs at ~30fps.
</blockquote></details>

Update: predicting irises now runs at ~20fps with the updated iris model that accepts batched inputs (we can now feed both eye patches to the model at once).
___
*[facemesh/src/index.ts, line 85 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/508#-MGk6PM60LSfy1v9eIrI:-MGsZawOAAh5Sxlk-5bj:b6g116h) ([raw file](https://github.com/tensorflow/tfjs-models/blob/15619fb7ebe32d0b23afa66a0ee79978f1fff071/facemesh/src/index.ts#L85)):*

How big is the iris model? should there be a flag for controlling if the irisModel should be loaded? Since user could be only using balzemesh model without iris detection.
</blockquote></details>

Done.

The iris model is 2.6MB. 
___
*[facemesh/src/index.ts, line 217 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/508#-MGk6yQwEM7lsrT6GGgU:-MGsat0iCeYJ1D0p-4KA:bjevr0n) ([raw file](https://github.com/tensorflow/tfjs-models/blob/15619fb7ebe32d0b23afa66a0ee79978f1fff071/facemesh/src/index.ts#L217)):*

should iris detection be default to false, since enabling it has a big performance hit?
</blockquote></details>

That's a tough call - with the new batched iris model the WebGL performance difference is 30fps (without iris) vs 20fps (with iris). I think the improvements from iris are worth the performance hit. I feel like if the user wants to predict facial geometry they would be willing to pay for accuracy - if they just want a bounding box they can always use blazeface instead. 

What do you think?

annxingyuan:
Hi @startupgurukul - thank you so much for taking the time to review the iris model and leave feedback. Regarding your suggestion to predict landmarks and irises in parallel - we actually need to predict landmarks first because the iris model expects a cropped image of the face as input, and we need the landmarks in order to know where to crop.

Regarding your suggestion to add a flag for predicting with / without iris - that's a great idea! I added a flag for both loading and predicting - you can disable loading the iris model by calling facemesh.load({shouldLoadIrisModel: false}), and at prediction time you can pass a parameter into facemesh.estimateFaces to avoid running the iris model. 

Regarding your suggestion to add an option to only return iris landmarks - since we have to find facial landmarks in order to run the iris model anyway, I think it makes sense to just have the face landmarks and face|iris landmarks options.

Thanks again and I hope you find the model useful once it's released :)

psw7420:
helo
i have run this project. but not running iris detect
even i click predictIrises checkbox, nothing shown. the result same with 'facemesh' 
the source code seems show iris by ellipse(), but nothing happend
what should i do to see my iris keypoint?

pyu10055 (approved):
> Reviewed 3 of 4 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/508)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @lina128, @pyu10055, and @tafsiri)

annxingyuan:
Hi @psw7420 - hmm that's strange - I've updated our public demo to include iris detection: https://storage.googleapis.com/tfjs-models/demos/facemesh/index.html

Do you see the iris keypoints there?

Skylion007:
@annxingyuan Works for me. 👍 How long till it's on npm? :D

annxingyuan:
@Skylion007 Hopefully within a week or so!

Maryasych:
> helo
> i have run this project. but not running iris detect
> even i click predictIrises checkbox, nothing shown. the result same with 'facemesh'
> the source code seems show iris by ellipse(), but nothing happend
> what should i do to see my iris keypoint?

same proplem.
https://storage.googleapis.com/tfjs-models/demos/facemesh/index.html works fine.

emannuell:
How can i solve, while facemesh 0.1.0 is not provided?
`Couldn't find any versions for "@tensorflow-models/facemesh" that matches "0.1.0"`

mhogg:
Hi,

I just tested this new facemesh with iris model from the facemesh_rotate_iris branch. Thanks for adding this. 

I have used face mesh previously, via the Augmented Faces example from Google ARCore in Unity, but I have been keen to try the new iris model and this javascript version seemed to be the easiest way. I was hoping that the iris model would provide better scaling of the face, resulting in more accurate face measurements. 

Firstly, I was able to run it this demo locally. I made some additions to index.js to take some measurements between mesh points. I have some questions:
* Compared to the Augmented Face model from Google ARCore (running on an Android mobile device), this javascript version gives different results. I was expecting the same results, so not sure why this would be. The javascript demo gives a similar measurement for nose width, but other measurements seem to be smaller than they should be.
* The face model prediction has a scaledMesh and mesh. The former has 478 points, but the latter only has 468. Why is this? Shouldn't mesh also have coordinates for the additional 10 points for the iris model?

My ultimate question is: If we know that the iris diameter is ~11.7mm, then shouldn't we be able to use this to improve the accuracy of the face mesh measurements when using the iris model, compared to the facemesh without the iris model? 

Thanks for your help,
Michael

annxingyuan:
Closed in favor of https://github.com/tensorflow/tfjs-models/pull/526

annxingyuan:
Hi @mhogg - thanks for trying out the model. In response to your questions:
- I'm not sure why the models aren't behaving identically - thanks for pointing this out. Though we started off using the same facemesh model, the postprocessing pipeline may have evolved slightly differently in ARCore vs TF.js. 
- Thanks for pointing out the discrepancy between scaledMesh and mesh! This should be resolved in the new PR: https://github.com/tensorflow/tfjs-models/pull/526
- Regarding using iris keypoints to refine facemesh measurements - that's a good point - we are already using iris keypoints to refine the contours around the eye, but perhaps they could be used to refine the entire mesh. I will investigate.

PR #510, opened 2020-08-17
[toxicity] Improve demo accessibility

eps1lon:
Clickable `<div>` are not accessible. With some minor CSS tweaks we can not only leverage a standard button but also a forms. That way pressing `Enter` while typing inside the input will add the entry. Before that you could only add an entry by clicking. Keyboard only was impossible.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/510)


eps1lon:
@googlebot I signed it!

annxingyuan (approved):
Thank you @eps1lon !

PR #516, opened 2020-08-28
Suggesting an e5 target for better compatibility

alangsto:
I'm suggesting that the target in the tsconfig.json file be es5 instead of es2017 for browser compatibility.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/516)

### DIFF
--- a/blazeface/tsconfig.json
+++ b/blazeface/tsconfig.json
@@ -7,7 +7,7 @@
     "removeComments": true,
     "preserveConstEnums": true,
     "declaration": true,
-    "target": "es2017",
+    "target": "es5",
     "lib": [
       "es2015",
       "dom"


tafsiri:
@annxingyuan this looks good to me, but i'm also wondering if there is a specific reason blazeface doesn't inherit from the top level tsconfig which sets the target lang to es5.

alangsto:
Also wanted to note that I forked this folder to test it with these changes and it is still working as expected.

alangsto:
@tafsiri I see that changes have been made to the blazeface/tsconfig.json file that appear to make this package es5 compatible. Is there anyway a new version of blazeface can be released that contains these changes?

tafsiri:
@annxingyuan do you have a blazeface update planned soon? If not maybe doing one to push this update would be good.

@alangsto if in the mean you need an es5 build you can load [blazeface.min.js](https://cdn.jsdelivr.net/npm/@tensorflow-models/blazeface@0.0.5/dist/blazeface.min.js) directly.

alangsto:
@tafsiri @annxingyuan are there any plans to publish a new blazeface version?

jenperson:
Thank you for your contribution! This PR is being closed because the package has been deprecated.

PR #530, opened 2020-10-15
[blazeface] Blazeface.load() takes optional modelURL param

jjs-stripe:
For commercial projects, the models generally need to be hosted on a trusted.& controlled site with a shared domain as the app. Adding a URL parameter lets projects store the model files on their own public servers.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/530)

### DIFF
--- a/blazeface/src/index.ts
+++ b/blazeface/src/index.ts
@@ -32,16 +32,19 @@ const BLAZEFACE_MODEL_URL =
  * much.
  *  `scoreThreshold` The threshold for deciding when to remove boxes based
  * on score.
+ *  `blazeModelUrl` Optional param for specifying a custom blaze model url or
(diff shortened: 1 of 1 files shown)


pyu10055:
Thank you for the contribution, the blaze face model has been deprecated, please use face-feature-detection model.

jjs-stripe:
@pyu10055 
We are using `face-landmarks-detection`, but that project itself loads `blazeface`. I have another PR that fixes `face-landmarks-detection` so that the models it loads can be specified via a parameter, but first I need to make blazeface be able to take a parameter

https://github.com/tensorflow/tfjs-models/blob/master/face-landmarks-detection/src/mediapipe-facemesh/index.ts#L153

PR #531, opened 2020-10-20
face-landmarks-detection: preserve aspect ratio

thebookins:
The bounding boxes used in face-landmarks detection are (in general) rectangular, whether they are derived from blazeface or from a previous iteration of landmark detection. These bounding boxes are then used to clip the input image and the resulting rectangular clip is resampled to a square of size 192 x 192 pixels before being input to the facemesh model.

This leads to a change in the aspect ratio of the image. The `prediction` object that is returned by the model contains a `mesh` object, which is affected by this change in aspect ratio, and a `scaledMesh` object which is not affected (the change in aspect ratio is reversed on projection back to the original image).

This appears to deviate from the MediaPipe implementation (see for example [this module](https://github.com/google/mediapipe/blob/master/mediapipe/modules/face_landmark/face_landmark_cpu.pbtxt) where "the scale_mode option is set to FIT to preserve the aspect ratio, resulting in potential letterboxing in the transformed image").

This change in aspect ratio means that the facemesh model is being input images with differential scaling in the vertical and horizontal directions, potentially leading to a degradation in model performance, as the images that were used to train the model were not scaled in this way. Furthermore, if iris detection is used, the iris model input images are cropped from the face clip, and share the differential vertical and horizontal scaling. The effect of this is evident on the `scaledMesh` object, where the detected irises are significantly elliptical (usually taller than they are wide).

This PR is an attempt to resolve this issue by squarifying the ROI boxes before clipping and resampling.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/531)


thebookins:
@googlebot I signed it!

annxingyuan (approved):
LGTM

wow thank you for catching this @thebookins ... I will update the demo ASAP :D

PR #534, opened 2020-10-21
[blazeface] Blazeface.load() takes optional modelURL param

jjs-stripe:
`mediapipe-facemesh` depends on `blazeface` for face detection.
However, blazeface does not allow loading models from non-tfhub sources. This is problematic for commercial users who do not want hard dependencies on other sites at web runtime.
This fix will allow [face-landmarks-detection](https://github.com/tensorflow/tfjs-models/blob/master/face-landmarks-detection/src/mediapipe-facemesh/index.ts#L153) to load blazeface from a custom URL.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/534)

### DIFF
--- a/blazeface/src/index.ts
+++ b/blazeface/src/index.ts
@@ -32,16 +32,19 @@ const BLAZEFACE_MODEL_URL =
  * much.
  *  `scoreThreshold` The threshold for deciding when to remove boxes based
  * on score.
+ *  `blazeModelUrl` Optional param for specifying a custom blaze model url or
+ * a `tf.io.IOHandler` object.
  */
 export async function load({
   maxFaces = 10,
   inputWidth = 128,
   inputHeight = 128,
   iouThreshold = 0.3,
-  scoreThreshold = 0.75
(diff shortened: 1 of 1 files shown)


tim-montague:
A baked in URL request is just bad practice in general.

1. It's not disclosed to users in the documentation
2. HTTP requests to outside URLs are security risks

The tensorflow model is simply a JSON file.

3. Users should be able to use their own HTTP client (Fetch, XHR, axios, etc) or a file system read (if using Node.js / Deno)

```
const model = fetch(URL).then(response => response.json());

Blazeface.load(model);
```

tim-montague:
I think the `blazeface.load()` feature is actually bloat and should be removed from the package. The documentation should instead provide an example with fetch (or Axios) to load the `https://tfhub.dev/tensorflow/tfjs-model/blazeface/1/default/1` JSON file.

pyu10055:
Another PR has been merged with the similar feature.

PR #537, opened 2020-11-22
[BodyPix] Fix image rendering only at natural scale in output renderer utility

amannm:
Hey guys, awesome project. Thanks for making it available.

This is a fix for the output renderer utility whose `drawMask` is currently ignoring scaling factors while writing the image to the canvas element, resulting in a misalignment between the image and the mask that's supposed to be superimposed on it. See https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage#Understanding_source_element_size

Below is an example of the issue from my test bench app, before and after recompiling the library with the fix:

Before:
Screen Shot 2020-11-22 at 9 15 16 AM

After:
Screen Shot 2020-11-22 at 9 14 05 AM

Thanks

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/537)

### DIFF
--- a/body-pix/src/output_rendering_util.ts
+++ b/body-pix/src/output_rendering_util.ts
@@ -363,7 +363,7 @@ export function drawMask(
     flipCanvasHorizontal(canvas);
   }
 
-  ctx.drawImage(image, 0, 0);
+  ctx.drawImage(image, 0, 0, width, height);
 
   ctx.globalAlpha = maskOpacity;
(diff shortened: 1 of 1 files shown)


amannm:
Whoops, looks like there's already a PR out for this one: https://github.com/tensorflow/tfjs-models/pull/517 I'll go ahead and close this one.

PR #543, opened 2020-12-03
Function signature overload proposal [body-pix]

magom001:
Hi! I would to make a humble contribution.

Currently `toMask` has default values for nearly all options. But in order to change `foregroundIds` for instance one has to provide values for all preceding arguments. As an example I overloaded `toMask` with alternative call signature.

If this approach is ok I am ready to overload other functions.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/543)


magom001:
@googlebot I signed it!

magom001:
@annxingyuan Hi! Is it possible to view why the build failed?

jenperson:
Thank you for your contribution! This PR is being closed because the package has been deprecated.

PR #547, opened 2020-12-14
Allow to provide custom blazeface url/io handler

SevInf:
Add `modelUrl` option to blazeface package, which allows to override
default url or provide custom IOHandler for loading the model.
Add `detectorModelUrl` option to `face-landmarks-detection` package, which
allows to customize blazeface url, used by mediapipe-facemesh.

Note, that the build for `face-landmarks-detection` fails in this branch. This is due to a fact that `face-landmarks-detection` depends on published version of `blazeface`, which does not have `modelUrl` parameter yet. Let me know if you'd prefer to have separate PRs for those 2 packages.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/547)


simon-lanf:
how do you use the tf.IO handler ? could you give us an example ? and can I ask which model you are loading ? I'm interested in testing different blazeface models

SevInf:
> how do you use the tf.IO handler ? could you give us an example ? and can I ask which model you are loading ? I'm interested in testing different blazeface models

In our case, we don't load a different model, we load the same model, but in a different way/from a different place. To be more specific, we want to bundle model's json config with a webpack and load weight files from webpack-generated asset urls to our CDN. We use a custom IOHandler to achieve this. If there is an interest, I can probably opensource it.

pyu10055 (approved):

PR #548, opened 2020-12-16
Remove chained op usage

lina128:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/548)

### DIFF
--- a/blazeface/package.json
+++ b/blazeface/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow-models/blazeface",
-  "version": "0.0.5",
+  "version": "0.0.6",
   "description": "Pretrained face detection model in TensorFlow.js",
   "main": "dist/index.js",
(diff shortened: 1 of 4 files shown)


annxingyuan (changes requested):
Thanks Na - did you want to also bump the dependencies in the blazeface/package.json to 2.8.0? You could also bump the version number for blazeface itself and then we can publish an update with these changes, then face-landmarks-detection can update its blazeface dependency.

lina128 (commented):
Bumped the version.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/548#-:-MOhvxgoEeA0-rnuZufp:b-rk1cmt)* status: 0 of 1 approvals obtained (waiting on @tafsiri)

annxingyuan (approved):

PR #552, opened 2021-01-05
Update package.json

rthadur:
Fixes https://github.com/tensorflow/tfjs/issues/4484

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/552)

### DIFF
--- a/body-pix/package.json
+++ b/body-pix/package.json
@@ -17,8 +17,8 @@
     "@tensorflow/tfjs-core": "^1.3.1"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


tafsiri:
@rthadur look like the tests are failing because the peerDependency is still at 1.3.1. Could you change that to match your update. Also could you run yarn in this directory to update the yarn.lock file for this model. Thanks.

rthadur:
@tafsiri sure will do, thank you

PR #568, opened 2021-01-21
Upgrade mobilenet to 3.0.0-rc.1

jinjingforever:
I also cleaned up the demo a bit to make it more like the other demos. Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/568)


tafsiri (commented):
Tests are currently failing but i think that can added by adding --skip-ignore. i.e. ts-node --skip-ignore --project tsconfig.test.json run_tests.ts,

Thanks for the overall cleanup!

> Reviewed 7 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/568#-:-MRagqxn9NppZRB-MrjB:b-e5o2v6)* status: 0 of 1 approvals obtained (waiting on @jinjingforever)

___
*[mobilenet/src/index.ts, line 20 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/568#-MRahKy28pzOzseGPlvK:-MRahKy28pzOzseGPlvL:b9pzs5b) ([raw file](https://github.com/tensorflow/tfjs-models/blob/69766d6a1e1a2cbbb6e1f4d0b0bc3fd3af237f7f/mobilenet/src/index.ts#L20)):*
> ```XML
> import * as tfconv from '@tensorflow/tfjs-converter';
> import '@tensorflow/tfjs-core/dist/public/chained_ops/add'
> ```

While this works, given the recent experience syncing to g3, I suspect we will move away from this pattern internally. We don't actually import this model into g3 though so its not necessary to do this right now. But I think in the long term our model code should just use the ops directly and this might be a good time to fix that. WDYT?

jinjingforever (commented):
Thanks Yannick! PTAL

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/568#-:-MRaw8dYE4okpDqu4ktq:bxn72v9)* status: 0 of 1 approvals obtained (waiting on @tafsiri)

___
*[mobilenet/src/index.ts, line 20 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/568#-MRahKy28pzOzseGPlvK:-MRaw6HR6NjuzeE1gSVt:by19uwp) ([raw file](https://github.com/tensorflow/tfjs-models/blob/69766d6a1e1a2cbbb6e1f4d0b0bc3fd3af237f7f/mobilenet/src/index.ts#L20)):*
<details><summary><i>Previously, tafsiri (Yannick Assogba) wrote…</i></summary><blockquote>

While this works, given the recent experience syncing to g3, I suspect we will move away from this pattern internally. We don't actually import this model into g3 though so its not necessary to do this right now. But I think in the long term our model code should just use the ops directly and this might be a good time to fix that. WDYT?
</blockquote></details>

Sounds good! Done.

jinjingforever:
hmm sorry, tests are failing. Looking into them..

jinjingforever:
OK. Now tests should be fixed. PTAL:) Thanks!

tafsiri (approved):
LGTM!

> Reviewed 1 of 9 files at r1, 1 of 2 files at r2, 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/568#-:-MRb-3KRCQhA8QHz9IoA:b-85i648)* status: :shipit: complete! 1 of 1 approvals obtained

PR #584, opened 2021-02-13
Adding a parameter to control the update behaviour of face bounding boxes from previous predictions on face-landmarks-detection

lmiguelmh:
Added `boundingBoxUpdateRoiIouThreshold` param to the load model function, which allows to decide how much does face detector bounding boxes overlap to be updated. This means that the model could be feed with images of different sizes and still work without needing to [force the model to invalidate the cache of face bounding boxes](https://github.com/tensorflow/tfjs/issues/4668#issue-805720307).

This change was suggested on the issue: [face-landmarks-detection model caching bounding boxes from previous predictions](https://github.com/tensorflow/tfjs/issues/4668). The local tests were successful as can be seen on these images used on the original issue: [facelandmarks-detection wrong when image dimensions change](https://github.com/tensorflow/tfjs/issues/4131).

```
    model = await faceLandmarksDetection.load(
      faceLandmarksDetection.SupportedPackages.mediapipeFacemesh,
      {
        maxContinuousChecks: 0,
        boundingBoxUpdateRoiIouThreshold: 1.0,  // update *all* face bounding boxes
      }
    );
    ...
    let predictions = await model.estimateFaces({
        input: imageAsTensor3D
      }
    );
    // draw landmarks and immediately predict another image of different size
    ...
    let predictions2 = await model.estimateFaces({
        input: image2AsTensor3D
      }
    );
    // draw landmarks
```

image

And this will be the result if using the default value of `0.25` for `boundingBoxUpdateRoiIouThreshold`:

image

**Note**. The build for face-landmarks-detection will fail in this branch because it depends on a published version of blazeface, which does not have modelUrl parameter yet.

### DIFF
--- a/face-landmarks-detection/src/index.ts
+++ b/face-landmarks-detection/src/index.ts
@@ -47,6 +47,8 @@ export enum SupportedPackages {
  * Defaults to 0.3.
  *  - `scoreThreshold` A threshold for deciding when to remove boxes based
  * on score in non-maximum suppression. Defaults to 0.75.
+ *  - `boundingBoxUpdateRoiIouThreshold` A threshold to decide how much does
+ *  face detector bounding boxes overlap to be updated. Defaults to 0.25.
  *  - `shouldLoadIrisModel` Whether to also load the iris detection model.
  * Defaults to true.
  */
--- a/face-landmarks-detection/src/mediapipe-facemesh/index.ts
+++ b/face-landmarks-detection/src/mediapipe-facemesh/index.ts
@@ -100,6 +100,8 @@ export type AnnotatedPrediction =
  * Defaults to 0.3.
  *  - `scoreThreshold` A threshold for deciding when to remove boxes based
  * on score in non-maximum suppression. Defaults to 0.75.
+ *  - `boundingBoxUpdateRoiIouThreshold` A threshold to decide how much does
+ *  face detector bounding boxes overlap to be updated. Defaults to 0.25.
  *  - `shouldLoadIrisModel` Whether to also load the iris detection model.
  * Defaults to true.
  *  - `modelUrl` Optional param for specifying a custom facemesh model url or
@@ -115,6 +117,7 @@ export async function load(config: {
   maxFaces?: number,
   iouThreshold?: number,
   scoreThreshold?: number,
+  boundingBoxUpdateRoiIouThreshold?: number,
   shouldLoadIrisModel?: boolean,
   modelUrl?: string|tf.io.IOHandler,
   detectorModelUrl?: string|tf.io.IOHandler,
@@ -126,6 +129,7 @@ export async function load(config: {
     maxFaces = 10,
     iouThreshold = 0.3,
     scoreThreshold = 0.75,
+    boundingBoxUpdateRoiIouThreshold = 0.25,
     shouldLoadIrisModel = true,
     modelUrl,
     detectorModelUrl,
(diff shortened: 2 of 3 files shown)


lmiguelmh:
@googlebot I signed it!

mattsoulanille:
Thanks @lmiguelmh for the PR! The bug this PR fixes is spread across a few issues, so I've summarized (my understanding of) it below.

In the face landmarks predictor, [regions of interest](https://github.com/tensorflow/tfjs-models/blob/master/face-landmarks-detection/src/mediapipe-facemesh/pipeline.ts#L114) determine where the predictor predicts face landmarks. Regions of interest are bounding boxes generated by Blazeface or by a prior run of the mesh / landmarks model. 

The first time the predictor is run, it needs to generate the regions of interest (bounding boxes around potential faces). It does so [by running Blazeface](https://github.com/tensorflow/tfjs-models/blob/master/face-landmarks-detection/src/mediapipe-facemesh/pipeline.ts#L254-L256) to detect faces and find bounding boxes for them. Then, it [runs the landmarks / mesh model on these bounding boxes](https://github.com/tensorflow/tfjs-models/blob/master/face-landmarks-detection/src/mediapipe-facemesh/pipeline.ts#L339-L341) to predict face landmarks. **It then [updates the positions of the (blazeface-generated) bounding boxes](https://github.com/tensorflow/tfjs-models/blob/master/face-landmarks-detection/src/mediapipe-facemesh/pipeline.ts#L411-L414) to the positions of the predicted face landmarks.** 

On subsequent runs, if the number of runs so far without Blazeface [hasn't exceeded `maxContinuousChecks`](https://github.com/tensorflow/tfjs-models/blob/master/face-landmarks-detection/src/mediapipe-facemesh/pipeline.ts#L480) the predictor may skip running Blazeface and use the bounding boxes generated from the mesh / landmarks model in the prior run. This improves performance since we're not running Blazeface and improves accuracy since the landmarks / mesh model tends to be more accurate at face position estimation than Blazeface (and thus produces better bounding boxes). If Blazeface _is_ run, then its bounding boxes are compared to the prior run's landmarks / mesh bounding boxes. If they differ less than an amount defined by [`UPDATE_REGION_OF_INTEREST_IOU_THRESHOLD`](https://github.com/tensorflow/tfjs-models/blob/master/face-landmarks-detection/src/mediapipe-facemesh/pipeline.ts#L34), then the [landmarks / mesh bounding boxes are used instead](https://github.com/tensorflow/tfjs-models/blob/master/face-landmarks-detection/src/mediapipe-facemesh/pipeline.ts#L430), since they tend to be more accurate especially for cases where the person's head is rotated or partially out of frame.

This technique relies on frames fed to the predictor being strongly correlated, such as in a video. To support uncorrelated predictions, we provide the variable `maxContinuousChecks` which defines how many times the predictor may run without re-running Blazeface to predict bounding boxes again. Setting this to zero causes Blazeface to be run for each prediction, which is desirable when predicting on independent still frames (we don't want to use old bounding boxes from an unrelated image). This is almost correct, except **we still compare Blazeface's bounding boxes to the prior run's landmarks / mesh bounding boxes and use the landmarks / mesh bounding boxes instead if they are similar enough**. This is incorrect since we're using bounding boxes from an unrelated frame, and it causes bad predictions, [especially when image size or aspect ratio changes](https://github.com/tensorflow/tfjs/issues/4668).

This PR allows the user to pass in a value for `UPDATE_REGION_OF_INTEREST_IOU_THRESHOLD`, letting them specify how similar the prior run's bounding boxes need to be to the current run's Blazeface bounding boxes for the prior run's bounding boxes to be used instead of the Blazeface bounding boxes. Passing zero requires an exact match, effectively using Blazeface's bounding boxes each time, which fixes the above bug in the case of unrelated input images.

Narendra186:
@lmiguelmh  Why isn't this PR merged? Is there any alternate solution?

PR #619, opened 2021-03-26
use video file  instead of camera ?

EugeneLiu2013:
@googlebot I signed it!

Anyone has use video file instead of camera ? estimate pose from video file could be very interesting

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/619)


jenperson:
This PR is being closed because the package is deprecated.

PR #626, opened 2021-03-31
Fix constructor call.

lina128:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/626)


ardoerlemans:
Thanks for the fix!

PR #628, opened 2021-04-01
[pose-detection]Fix lint

lina128:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/628)


pose-detection/package.json
```diff
@@ -34,6 +34,7 @@
     "rollup-plugin-visualizer": "~3.3.2",
     "ts-node": "~8.8.2",
     "tslint": "~5.18.0",
+    "tslint-no-circular-imports": "~0.5.0",
```
pyu10055, line 37: you should upgrade tslint and circular imports to the latest version to avoid security vulnerabilities.

pyu10055 (approved):

PR #658, opened 2021-04-19
[pose-detection]Change Blazepose fullbody and upperbody selection fro…

lina128:
This PR changes Blazepose fullbody and upperbody selection from a modelConfig option to enum option, which means fullbody and upperbody are two models. This is because the two models output different number of keypoints. There are code logic that is return keypoint array/map based on model (see util.ts). It is cleaner to decide on which keypoint representation just base on the model (enum) as oppose to basing on both model (enum) and config.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/658)

### DIFF
--- a/pose-detection/demo/src/index.js
+++ b/pose-detection/demo/src/index.js
@@ -37,9 +37,9 @@ async function createDetector() {
         inputResolution: {width: 500, height: 500},
         multiplier: 0.75
       });
-    case posedetection.SupportedModels.MediapipeBlazepose:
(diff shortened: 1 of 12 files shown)


pyu10055 (approved):
> Reviewed 12 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/658)* status: :shipit: complete! 1 of 1 approvals obtained

PR #656, opened 2021-04-19
[pose-detection]Use url querystring to control which model to load.

lina128:
Possible query strings:
1. ?model=posenet
2. ?model=blazepose
3. ?model=movenet
4. ?model=movenet&&type=thunder
5. ?model=movenet&&type=lightning

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/656)

PR #660, opened 2021-04-20
[pose-detection]Add model comparison to demo.

lina128:
In previous https://github.com/tensorflow/tfjs-models/pull/656, we use url to control which model to load and eliminate other models from showing in the UI. In this PR, we add back other model selection so that user can easily switch models for comparison. Initial loading is still controlled by url query string.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/660)

### DIFF
--- a/pose-detection/demo/src/camera.js
+++ b/pose-detection/demo/src/camera.js
@@ -104,7 +104,7 @@ export class Camera {
    */
   drawKeypoints(keypoints) {
     const keypointInd =
-        posedetection.util.getKeypointIndexBySide(params.STATE.model.model);
+        posedetection.util.getKeypointIndexBySide(params.STATE.model);
     this.ctx.fillStyle = 'White';
     this.ctx.strokeStyle = 'White';
     this.ctx.lineWidth = params.DEFAULT_LINE_WIDTH;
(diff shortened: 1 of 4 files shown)


pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/660)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

___
*[pose-detection/demo/src/option_panel.js, line 53 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/660#-MYlfqid6pmEfe7Ni1FQ:-MYlfqid6pmEfe7Ni1FR:b-94yebw) ([raw file](https://github.com/tensorflow/tfjs-models/blob/90163486bc2662f90b12ba8d8af97b612a147ded/pose-detection/demo/src/option_panel.js#L53)):*
> ```JavaScript
>       if (type !== 'lightning' && type !== 'thunder') {
>         // Nulify invalid value.
>         type = null;
> ```

should the type be initialized with a default value? same for blazepose case.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/660)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

___
*[pose-detection/demo/src/option_panel.js, line 53 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/660#-MYlfqid6pmEfe7Ni1FQ:-MYlpuHN9SGOuCZ1a2MS:bflu98q) ([raw file](https://github.com/tensorflow/tfjs-models/blob/90163486bc2662f90b12ba8d8af97b612a147ded/pose-detection/demo/src/option_panel.js#L53)):*

should the type be initialized with a default value? same for blazepose case.
</blockquote></details>

They will get a default in the method after this block, in updateModelConfigFolder(). In that method, null value will always get a default. Therefore, we don't need to repeat that logic here.

PR #663, opened 2021-04-22
[pose-detection]Add backend options to demo.

lina128:
This PR adds backend options (similar to our benchmark tool) to the demo, so that it allows us to easily switch optimization flags and test the result with real video.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/663)

### DIFF
--- a/pose-detection/demo/index.html
+++ b/pose-detection/demo/index.html
@@ -16,12 +16,10 @@ limitations under the License.
   body {
     margin: 25px;
   }
-
   #main {
     position: relative;
     margin: 50px 0;
   }
-
   canvas {
     position: absolute;
     top: 0;
--- a/pose-detection/demo/index.html
+++ b/pose-detection/demo/index.html
@@ -9,26 +9,26 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================-->
-
-<meta charset="utf-8">
-<meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1.0, user-scalable=no">
-<style>
-  body {
-    box-sizing: border-box;
-    font-family: 'Roboto', sans-serif;
-    font-size: 13px;
-    margin: 20px 100px;
-  }
-  #main {
-    position: relative;
-    margin: 50px 0;
-  }
-  canvas {
-    position: absolute;
-    top: 0;
-    left: 0;
-  }
-</style>
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1.0, user-scalable=no">
+  <style>
+    body {
+      margin: 25px;
+    }
+    #main {
+      position: relative;
+      margin: 50px 0;
+    }
+    canvas {
+      position: absolute;
+      top: 0;
+      left: 0;
+    }
+  </style>
+</head>
 <body>
   <div id="main">
     <div class="container">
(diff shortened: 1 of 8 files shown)


pyu10055 (changes requested):
This looks very nice, thank you for adding this. As we discussed offline, the backend and flag logic would be better to packaged in a separate npm, so all model demos and benchmark tool can share.
For example, the WEBGL_PACK_DEPTHWISECONV is no longer tunable, since it depends on WEBGL_PACK flag now.
Now it requires two places to update the logic. 
The other question is the version, some of the flags only works for certain TFJS version, the demo might using a version that is not compatible with the flags. Which means the new package should have the same version number, or be part of the union package.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/663#-:-MYrFR4V2uQbSj4T1iPF:b-s7k4hx)* status: 0 of 1 approvals obtained (waiting on @lina128)

___
*[pose-detection/demo/src/option_panel.js, line 86 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/663#-MYrH1bE2kA2TLptku6W:-MYrH1bE2kA2TLptku6X:b-h5jgmw) ([raw file](https://github.com/tensorflow/tfjs-models/blob/e9ed72fba7433a0491a96822b7aead7ef4532a95/pose-detection/demo/src/option_panel.js#L86)):*
> ```JavaScript
>   modelController.onChange(model => {
>     params.STATE.changeToModel = model;
>     showModelConfigs(modelFolder);
> ```

should the modelConfigFolder.open() still be called here?
___
*[pose-detection/demo/src/option_panel.js, line 93 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/663#-MYrHLag8SHMRBoAfacI:-MYrHLag8SHMRBoAfacJ:bsozqgy) ([raw file](https://github.com/tensorflow/tfjs-models/blob/e9ed72fba7433a0491a96822b7aead7ef4532a95/pose-detection/demo/src/option_panel.js#L93)):*
> ```JavaScript
>   modelFolder.open();
> 
>   const backendFolder = gui.addFolder('Backend');
> ```

might be good to create a method showBackend(model) that can conditionally create backends that are supported by the model.
This would be useful for MP web solution case.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/663)* status: 0 of 1 approvals obtained (waiting on @pyu10055)

___
*[pose-detection/demo/src/option_panel.js, line 86 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/663#-MYrH1bE2kA2TLptku6W:-MYul-wi692sY-WcJ6bS:b-2u20kr) ([raw file](https://github.com/tensorflow/tfjs-models/blob/e9ed72fba7433a0491a96822b7aead7ef4532a95/pose-detection/demo/src/option_panel.js#L86)):*

should the modelConfigFolder.open() still be called here?
</blockquote></details>

We remove the modelConfigFolder, so there's only one folder for all the model related options, i.e. modelFolder. The first option is model selection, the rest is dynamic, based on the first option. So we don't need to reopen the folder here again.
___
*[pose-detection/demo/src/option_panel.js, line 93 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/663#-MYrHLag8SHMRBoAfacI:-MYulVLb-85jcm1HZVcL:b7ehhjr) ([raw file](https://github.com/tensorflow/tfjs-models/blob/e9ed72fba7433a0491a96822b7aead7ef4532a95/pose-detection/demo/src/option_panel.js#L93)):*

might be good to create a method showBackend(model) that can conditionally create backends that are supported by the model.
This would be useful for MP web solution case.
</blockquote></details>

Good suggestion. There might be both selection for runtime and backends within the runtime. We'll see how it should be setup once MP is integrated.

pyu10055 (approved):
> Reviewed 5 of 8 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/663)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pyu10055)

PR #665, opened 2021-04-23
[pose-detection] Fix demo iPhone issue.

lina128:
There are two issues:
1. Sometimes the loadeddata is not ready when inference, so added code make sure to always wait for data is ready.
2. Large canvas rendering will freeze iPhone Safari, so added code to default to a smaller size if mobile devices. Also added the size option so that desktop can test it too.

Also add below fixes:
1. Update to use tfjs 3.5.0
2. Fix backend change logic.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/665)

### DIFF
--- a/pose-detection/demo/index.html
+++ b/pose-detection/demo/index.html
@@ -16,12 +16,10 @@ limitations under the License.
   body {
     margin: 25px;
   }
-
   #main {
     position: relative;
     margin: 50px 0;
   }
-
   canvas {
     position: absolute;
     top: 0;
--- a/pose-detection/demo/index.html
+++ b/pose-detection/demo/index.html
@@ -9,26 +9,26 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================-->
-
-<meta charset="utf-8">
-<meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1.0, user-scalable=no">
-<style>
-  body {
-    box-sizing: border-box;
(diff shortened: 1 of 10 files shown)


pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/665)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

___
*[pose-detection/demo/src/params.js, line 24 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/665#-MYwkzyh5Nb33kNOApgs:-MYwkzyiBd58Z-AUt6kZ:b-n9zb6r) ([raw file](https://github.com/tensorflow/tfjs-models/blob/c097653c194b542be02e8fe3db3efd74f6732082/pose-detection/demo/src/params.js#L24)):*
> ```JavaScript
> export const VIDEO_SIZE = {
>   '640 X 480': {width: 640, height: 480},
>   '640 X 360': {width: 640, height: 360},
> ```

Will the different aspect ratio 4:3, 16:9 affects the keypoint scaling?
___
*[pose-detection/demo/src/util.js, line 37 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/665#-MYwlqtj4HUqh9Yw1bQz:-MYwlqtj4HUqh9Yw1bR-:b-ddjztm) ([raw file](https://github.com/tensorflow/tfjs-models/blob/c097653c194b542be02e8fe3db3efd74f6732082/pose-detection/demo/src/util.js#L37)):*
> ```JavaScript
>   }
> 
>   const currentBackend = tf.getBackend();
> ```

currentBackend is not needed anymore?

PR #668, opened 2021-04-26
[pose-detection]Refactor code.

lina128:
This PR makes state update logic simpler. Instead of passing along changed value, it simply sets corresponding value change flag to true/false. When need to read the value, we can read directly from the value field, this is because onChange/onFinishChange event is fired after the value field is updated to new value. This refactor ensures:
(1) Value field is the only source of truth.
(2) All the value change flag fields are boolean, so that logic that are based on them can treat them as boolean without guessing. (Otherwise, some flag contains object, so need to check non-null, some flag contains boolean, so need to check truthy, etc.)

### DIFF
--- a/pose-detection/demo/src/index.js
+++ b/pose-detection/demo/src/index.js
@@ -56,26 +56,26 @@ async function createDetector() {
 }
 
 async function checkGuiUpdate() {
-  if (STATE.changeToTargetFPS || STATE.changeToSizeOption) {
+  if (STATE.isTargetFPSChanged || STATE.isSizeOptionChanged) {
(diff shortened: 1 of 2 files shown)


pyu10055 (approved):
Thanks for refactoring, if the logic becomes more complicated, might consider using the redux state management paradigm.

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/668#-:-MZEEIaV9QucSUMuKG70:b-kca3c5)* status: :shipit: complete! 1 of 1 approvals obtained

PR #682, opened 2021-04-29
[pose-detection] Make internal MoveNet method synchronous

ardoerlemans:
If this internal call is asynchronous, it causes issues with the crop region calculations if a new estimatePoses call is started before the previous one finishes.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/682)


lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/682)* status: 0 of 1 approvals obtained (waiting on @ardoerlemans)

___
*[pose-detection/src/movenet/detector.ts, line 124 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/682#-MZV25Kr6gm586DlfBl3:-MZV25Kr6gm586DlfBl4:b-pbhxlh) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9e5393aeb7951062820ce1ef85d4f7f9f1368502/pose-detection/src/movenet/detector.ts#L124)):*
> ```XML
>     }
> 
>     const inferenceResult = outputTensor.dataSync();
> ```

Can you still use await data() here?

ardoerlemans (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/682)* status: 0 of 1 approvals obtained (waiting on @lina128)

___
*[pose-detection/src/movenet/detector.ts, line 124 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/682#-MZV25Kr6gm586DlfBl3:-MZV2TI_7kwdcg5Tkhgg:b-i70sls) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9e5393aeb7951062820ce1ef85d4f7f9f1368502/pose-detection/src/movenet/detector.ts#L124)):*

Can you still use await data() here?
</blockquote></details>

Now that detectKeypoints is not async anymore, we're not allowed to use await in this method.

PR #683, opened 2021-04-29
TFJS task API initial version

jinjingforever:
This version is based on our discussion and the design doc. Hopefully the comments in the code make sense. Maybe start from `task_model.ts` to get a better understanding of the API structure. 

This PR also uses the API to define three task models: mobilenet (tfjs), mobilenet (tflite), and custom image classification tflite model. All dependent packages are dynamically loaded on demand (this package itself is only 9.8k minified without bundling any models). Same package from different models will only be loaded once. For tfjs model, you can specify which backend to use in the options. One web app can even have instances of the same model with different backends at the same time (backend switching is taken care of internally).

I made a [simple demo](https://codepen.io/jinjingforever/pen/VwPOePq) to show how things are used in practice. You can check from the network tab and see how packages are loaded.

(Note that this package is located in the tfjs-models repo. So any dependencies on the tfjs repo has to be pinned on a version instead of using relative paths)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683)


jinjingforever:
BTW I will add tests and hook them up with cloud build once this PR is merged:)

pyu10055 (changes requested):
Great work Jing, I took a pass and added couple comments, feel free to chat with me if you have questions.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-:-MZYhin5E9OZiKPxSRQU:b-ou4phe)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @mattsoulanille)

___
*[tasks/README.md, line 8 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYRuemAS6au3HchnVd:-MZYRuen5HJcuRCzIBhR:b-9s0tk2) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L8)):*
> ```Markdown
> on the web. It covers popular machine learning tasks, such as sentiment
> detection, image classification, pose detection, etc. For each task, you can
> choose from various models supported by different runtime systems, such as
> ```

choose from a collection of SoTA models regardless of the underlying runtime systems (i.e. TFJS, TFLite, MediaPipe etc).
___
*[tasks/README.md, line 9 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYSGWB2fc9mCdFkC0s:-MZYSGWB2fc9mCdFkC0t:bsibtci) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L9)):*
> ```Markdown
> detection, image classification, pose detection, etc. For each task, you can
> choose from various models supported by different runtime systems, such as
> TFJS, TFLite, MediaPipe, etc. The model interfaces are specifically designed
> ```

We are aiming to provide an intuitive and easy-to-use API for each task and improve usability for developers without ML knowledge.
___
*[tasks/README.md, line 46 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYT4CPCWZd9ipv-2kk:-MZYT4CPCWZd9ipv-2kl:baieh0x) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L46)):*
> ```Markdown
> 
> ```js
> import * as tftask from '@tensorflow-models/tasks';
> ```

how about using namespace as `tfTask`?
___
*[tasks/README.md, line 61 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYTJ8nBgR6CY1QTQpl:-MZYTJ8nBgR6CY1QTQpm:bl5hrtr) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L61)):*
> ```Markdown
> 
> Since all these models are for the `Image Classification` task, they will have
> the same task model type: `ImageClassifier` in this case. Each task model
> ```

it would be good provide a link to the ImageClassifier interface or doc when it is available.
___
*[tasks/README.md, line 62 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYTYHE6cuUmxezzr_7:-MZYTYHE6cuUmxezzr_8:b-vp59ty) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L62)):*
> ```Markdown
> Since all these models are for the `Image Classification` task, they will have
> the same task model type: `ImageClassifier` in this case. Each task model
> defines an unique and easy-to-use inference method that fits its task the best.
> ```

remove `that fits its task the best.`
___
*[tasks/README.md, line 78 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYTeuv3T-SalSbALpC:-MZYTeuv3T-SalSbALpD:benv1dj) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L78)):*
> ```Markdown
> metadata requirements of the corresonding task in the TFLite task library.
> Check out the "model compatibility requirements" section of the official task
> library page. For example, the requirements of `ImageClassifier` can be found
> ```

it would be good to provide a javascript example here.
___
*[tasks/rollup.config.js, line 66 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZW7-ge6oTUlXnl-u_E:-MZW7-gf2ZTzTbuapYF9:b1c37pg) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/rollup.config.js#L66)):*
> ```JavaScript
>   config({
>     plugins: [minify()],
>     output: {format: 'umd', name: packageName, file: 'dist/tfjs-tasks.min.js'}
> ```

should we add support es2017, node, and fesm outputs.
___
*[tasks/src/package_loader.ts, line 38 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYbK2dDXMawEndjZtR:-MZYbK2e-KsHQIZ6aCQA:bf8b9bj) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L38)):*
> ```XML
>    * url2 are both done loading.
>    */
>   async loadPackagesAndGetGlobalNs(
> ```

should the GlobalNamespaces be fully spelled out?
___
*[tasks/src/package_loader.ts, line 41 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZY_DWI2g_ZnX-5HESQ:-MZY_DWI2g_ZnX-5HESR:b820qoj) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L41)):*
> ```XML
>       namespace: string, packageUrls: Array<string[]>,
>       curIndex = 0): Promise<G> {
>     if (curIndex >= packageUrls.length) {
> ```

should it be `curIndex < packageUrls.length` ?
___
*[tasks/src/package_loader.ts, line 43 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYWeut9tlNRUG5AQBn:-MZYWeut9tlNRUG5AQBo:b-yjgk3b) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L43)):*
> ```XML
>     if (curIndex >= packageUrls.length) {
>       // tslint:disable-next-line:no-any
>       const global: any = isWebWorker() ? self : window;
> ```

might be good to follow tfjs convention that uses a platform class to allow customization of platform specific features.
For example fetch, global, etc that exist in browser not in node.
might be able to use tfjs env().platform method directly.
___
*[tasks/src/package_loader.ts, line 61 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYaESCCe4D1b70LBUk:-MZYaESCCe4D1b70LBUl:bqb2xu9) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L61)):*
> ```XML
>     // From webworker.
>     if (typeof window === 'undefined') {
>       importScripts(packageUrl);
> ```

will importScripts be more efficient to load multiple urls together?
___
*[tasks/src/task_model.ts, line 94 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYb-BOA5Q3qLNPvSIH:-MZYb-BOA5Q3qLNPvSII:b-ohe362) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/task_model.ts#L94)):*
> ```XML
> export abstract class TaskModelLoader<N, LO, M> {
>   /** The model name. */
>   abstract readonly name: string;
> ```

It would be good to have a single metadata interface that contains all of the information like name, description, ...
___
*[tasks/src/tasks/common.ts, line 41 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYfnXp8yNpqj7d8qj6:-MZYfnXp8yNpqj7d8qj7:bcwo3bm) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/common.ts#L41)):*
> ```XML
> /** All supported tasks. */
> export enum Task {
>   IMAGE_CLASSIFICATION = 'IMAGE_CLASSIFICATION',
> ```

if the name and string is the same, I think you can omit the assignment.
___
*[tasks/src/tasks/common.ts, line 76 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYg9UkB6LUQG_ZWdLn:-MZYg9UkB6LUQG_ZWdLo:bhsqxn7) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/common.ts#L76)):*
> ```XML
>       packages[1].push(
>           `https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm@${
>               version}/dist/tf-backend-wasm.js`);
> ```

don't need to specify the un-minimized file?
___
*[tasks/src/tasks/image_classification/custom_model_tflite.ts, line 58 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYgp6HAv3qBCmkynwl:-MZYgp6I1XiGEhpw_MwJ:bu7t3ln) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/image_classification/custom_model_tflite.ts#L58)):*
> ```XML
>   readonly packageUrls =
>       [[`https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-tflite@${
>           this.version}/dist/tf-tflite.min.js`]];
> ```

I think the portion of `/dist/tf-tflite.min.js` is by default.
___
*[tasks/src/tasks/image_classification/mobilenet_tfjs.ts, line 79 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYhaej8xY8bw-oOE8k:-MZYhaej8xY8bw-oOE8l:b-ri9n8z) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/image_classification/mobilenet_tfjs.ts#L79)):*
> ```XML
>       },
> 
>       cleanUp: () => {},
> ```

I think tfjs mobilenet should have a dispose() method where it releases the model

pyu10055 (commented):
BTW the demo is simply brilliant! @jasonmayes 

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-:-MZYi7q_3i2vEvYtFMcj:bhs82lo)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @mattsoulanille)

jinjingforever (commented):
Thanks so much for the review! PTAL

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-:-MZZ02HkDT8uX-w2IsaM:bdetmh9)* status: 0 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, and @pyu10055)

___
*[tasks/README.md, line 8 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYRuemAS6au3HchnVd:-MZYoAVy0h3QK9TD9Pai:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L8)):*

choose from a collection of SoTA models regardless of the underlying runtime systems (i.e. TFJS, TFLite, MediaPipe etc).
</blockquote></details>

Done.
___
*[tasks/README.md, line 9 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYSGWB2fc9mCdFkC0s:-MZYoAtK-GNwzQ4dj6D7:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L9)):*

We are aiming to provide an intuitive and easy-to-use API for each task and improve usability for developers without ML knowledge.
</blockquote></details>

Done.
___
*[tasks/README.md, line 46 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYT4CPCWZd9ipv-2kk:-MZYoKZz90ISbq4Qj1e3:b5c148w) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L46)):*

how about using namespace as `tfTask`?
</blockquote></details>

sgtm! Also updated the rollup config.
___
*[tasks/README.md, line 61 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYTJ8nBgR6CY1QTQpl:-MZYp5j-B9cqKYwtIS2t:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L61)):*

it would be good provide a link to the ImageClassifier interface or doc when it is available.
</blockquote></details>

Done.
___
*[tasks/README.md, line 62 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYTYHE6cuUmxezzr_7:-MZYpEPnDuB1XiyyUCOL:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L62)):*

remove `that fits its task the best.`
</blockquote></details>

Done.
___
*[tasks/README.md, line 78 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYTeuv3T-SalSbALpC:-MZYpUI01HeisaWCxrTI:b5jlbg5) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L78)):*

it would be good to provide a javascript example here.
</blockquote></details>

Are you talking about an example for custom tflite model? If so, there is one above in the "load model and run inference" section.
___
*[tasks/rollup.config.js, line 66 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZW7-ge6oTUlXnl-u_E:-MZYrit29dsiXZrfd3_K:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/rollup.config.js#L66)):*

should we add support es2017, node, and fesm outputs.
</blockquote></details>

Done.
___
*[tasks/src/package_loader.ts, line 38 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYbK2dDXMawEndjZtR:-MZYs4dK4Y0fBpIJlYkZ:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L38)):*

should the GlobalNamespaces be fully spelled out?
</blockquote></details>

Done.
___
*[tasks/src/package_loader.ts, line 41 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZY_DWI2g_ZnX-5HESQ:-MZYsBBXAcz9hxKvsYxd:b-og4d7x) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L41)):*

should it be `curIndex < packageUrls.length` ?
</blockquote></details>

This is the termination condition of the recursive calls. It basically means if there is no more packages to fetch, return the global namespace.
___
*[tasks/src/package_loader.ts, line 43 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYWeut9tlNRUG5AQBn:-MZYuSQlEp5wjmwUj_Z-:b-u9kxp2) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L43)):*

might be good to follow tfjs convention that uses a platform class to allow customization of platform specific features.
For example fetch, global, etc that exist in browser not in node.
might be able to use tfjs env().platform method directly.
</blockquote></details>

Good idea! I took a look at env().platform and it looks like it provides different implementations for "fetch" on different platforms. We are not actually using fetch here (in theory we could use fetch+eval to execute javascript but it is not recommended).

But using a similar design here is a great idea. I will add a TODO here for now. I will get to that when I start testing this in nodejs (after IO), especially the dynamic loading part. I am adding a sentence in README to say that the initial version only supports browser. NodeJS support is coming soon.  
___
*[tasks/src/package_loader.ts, line 61 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYaESCCe4D1b70LBUk:-MZYwtrRC9-oD_95Tmmm:b-adr56d) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L61)):*

will importScripts be more efficient to load multiple urls together?
</blockquote></details>

Looks like importScripts actually imports multiple scripts in sequence, not in parallel
___
*[tasks/src/task_model.ts, line 94 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYb-BOA5Q3qLNPvSIH:-MZYylhz3gHT_BnwgNqk:b2ifwk0) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/task_model.ts#L94)):*

It would be good to have a single metadata interface that contains all of the information like name, description, ...
</blockquote></details>

Great idea! Done
___
*[tasks/src/tasks/common.ts, line 41 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYfnXp8yNpqj7d8qj6:-MZYz-HA-82NE797cOtG:buo95za) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/common.ts#L41)):*

if the name and string is the same, I think you can omit the assignment.
</blockquote></details>

This is intentional because I want it to be string enum so I can use them in the model loader index as keys.
___
*[tasks/src/tasks/common.ts, line 76 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYg9UkB6LUQG_ZWdLn:-MZYzSAC-_UDxqwb8oFJ:b-to5bq) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/common.ts#L76)):*
> ```
> https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm
> ```

Actually for wasm backend to work, I will have to load tf-backend-wasm script explicitly. If I load "https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm", it will try to load the wasm files from "https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm/" instead of "https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm/dist/" (that is without using setWasmPath).

But you reminded me that I should use minified file here:) Updated.
___
*[tasks/src/tasks/image_classification/custom_model_tflite.ts, line 58 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYgp6HAv3qBCmkynwl:-MZZ-EcMEO6pvvTA_AwH:b-sgz9fr) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/image_classification/custom_model_tflite.ts#L58)):*

I think the portion of `/dist/tf-tflite.min.js` is by default.
</blockquote></details>

See comments above about wasm loading path.
___
*[tasks/src/tasks/image_classification/mobilenet_tfjs.ts, line 79 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYhaej8xY8bw-oOE8k:-MZZ-YpHEsq9P4YS4HNF:bjs77dt) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/image_classification/mobilenet_tfjs.ts#L79)):*

I think tfjs mobilenet should have a dispose() method where it releases the model
</blockquote></details>

Looks like it disposes stuff internally. I don't see any public facing dispose method. https://github.com/tensorflow/tfjs-models/blob/master/mobilenet/src/index.ts#L145

lina128 (commented):
Great work, clean API! Only left minor comments, see below.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-:-MZZIrC60S3zOY-gvDmp:b-xm3i55)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @mattsoulanille, and @pyu10055)

___
*[tasks/README.md, line 6 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYzGQ632xdVqH6I4_N:-MZYzGQ632xdVqH6I4_O:bnrinkv) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L6)):*
> ```
> web
> ```
Web
___
*[tasks/README.md, line 9 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYzODI3VkJr6FqrsHw:-MZYzODI3VkJr6FqrsHx:b-720obf) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L9)):*
> ```
> TFJS
> ```
Add link to GitHub.
___
*[tasks/README.md, line 9 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYzSEEA9ru0UB-I2H7:-MZYzSEF6Z-WadsbQ5ng:b-2idq1i) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L9)):*
> ```
> TFLite
> ```
Add link to GitHub.
___
*[tasks/README.md, line 9 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYzTIQ1-vK37WqpUV8:-MZYzTIQ1-vK37WqpUV9:bdyfz06) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L9)):*
> ```
> MediaPipe
> ```
Add link to GitHub.
___
*[tasks/README.md, line 10 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYzZcO2jKyliSMPr1b:-MZYzZcPABvbWynxWdve:b61ebaf) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L10)):*
> ```
> The model interfaces are specifically designed
> for each task. 
> ```
This sentence is confusing, hard to understand what message it's trying to convey. Consider removing or rephrase.
___
*[tasks/README.md, line 11 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ-7ae1GaBiq93fouz:-MZZ-7ae1GaBiq93fov-:b-dclls6) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L11)):*
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>

> ```
> for each task. They are intuitive and easy-to-use, even for developers without
> any comprehensive ML knowledge. In addition, the API will automatically load
> required packages on the fly. You never need to worry about missing dependencies
> again.
> ```
</details>

I understand this part and it is great, but for first time reader, it's hard to follow all those points and the messages are not very clear. Try to write more constructively. What about something like:
The Task API is designed with ease-of-use in mind, it has below features:
1. Models are grouped by ML tasks, making them easy to discover.
2. Simple installation, you only need to install this API without needing to worry about all the models, runtimes, backends.
3. Lightweight package, only 9.8kb. Models, runtimes and backends are only dynamically loaded when needed, and will only be loaded once per page load.
___
*[tasks/README.md, line 63 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ259F6yVrv8pf7me2:-MZZ259F6yVrv8pf7me3:b-h8xy55) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L63)):*
> ```
> classify
> ```
Is it possible to simplify the API to `load`, `predict` for all task models? Using `classify` in one task and `estimate` in another task basically have the same effect, which is run prediction, right? TFLite task API uses `classify`, `estimate` etc., so does our tfjs-models. ML5.js uses `predict` for every task. MediaPipe uses `send` for every task. So there are examples in both leagues, but personally I prefer the latter, just a simple API, easy to use, easy to remember, for any users.
___
*[tasks/README.md, line 86 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ3FNw1raSmgF_sFcr:-MZZ3FNw1raSmgF_sFcs:but22fe) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L86)):*
> ```Markdown
> For most cases, the WebGL backend (default) is usually the fastest.
> 
> For TFLite models, we use WebAssembly under the hood. It uses [XNNPACK][xnnpack]
> ```

Some of the contents here are pretty advanced, it is contrary to what said above, for users who are not familiar with ML. Maybe separate the contents with a notice or section head, 'More Advanced Contents for Eager Readers'.
___
*[tasks/src/package_loader.ts, line 30 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ5UDP-zpfhXWAf-V7:-MZZ5UDP-zpfhXWAf-V8:b287qw2) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L30)):*
> ```XML
>    * Loads the given packages and returns the loaded global namespace.
>    *
>    * Each element in `packageUrls` is an array of package urls (package set).
> ```

Before the detailed description, it's better to summarize packageUrls allows you to express package dependency, which determines loading order.
___
*[tasks/src/package_loader.ts, line 36 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ54rR71xjtYWqx6ZG:-MZZ54rR71xjtYWqx6ZH:b-5z7xj) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L36)):*
> ```XML
>    * For example, if packageUrls = [ [url1, url2], [url3] ], then url1 and
>    * url2 are loaded together first. url3 will then be loaded when url1 and
>    * url2 are both done loading.
> ```

It's better to add somewhere here that this method allows multi-level nested arrays, such as [[[url1, url2], url3], url4]
___
*[tasks/src/package_loader.ts, line 44 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ5vTu8QBOuGLSi8bW:-MZZ5vTu8QBOuGLSi8bX:b-xf03wu) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L44)):*
> ```XML
>       // tslint:disable-next-line:no-any
>       const global: any = isWebWorker() ? self : window;
>       return global[namespace] as G;
> ```

Does it handle namespace conflict? What if namespace already exists?
___
*[tasks/src/package_loader.ts, line 66 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ8sfqC5UUY5jEaMUu:-MZZ8sfqC5UUY5jEaMUv:b-gf18kh) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L66)):*
> ```XML
>     // From webpage.
>     else {
>       const script = document.createElement('script');
> ```

You've researched security issue, right? Any known risk? Do we need to mention it in the README?
___
*[tasks/src/package_loader.ts, line 38 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZBoQNCdw9N4D85DyI:-MZZBoQNCdw9N4D85DyJ:b-ywwiz1) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/package_loader.ts#L38)):*
> ```XML
>    * url2 are both done loading.
>    *
>    * TODO: use a design similar to tfjs' `Platform` to provide different
> ```

This doccomment doesn't have parameter description, please add. What is the naming convention for namespace?
___
*[tasks/src/task_model.ts, line 72 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ9sLW6Tmk8fmcDgpQ:-MZZ9sLW6Tmk8fmcDgpR:b-ugm4jh) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/task_model.ts#L72)):*
> ```
>  so it makes more sense to put them here
> ```
This is design details, no need to expose it in doccomment.
___
*[tasks/src/task_model.ts, line 59 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZJqBXCt98yYEbihrv:-MZZJqBXCt98yYEbihrw:b-e0b9s5) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L59)):*
> ```XML
>    * Cleans up resources if necessary.
>    */
>   cleanUp(): void;
> ```

Some pipelines have global states (all MP models and pose detection models), and has a method to reset states, how to support that? Either ask user to clean up and recreate an instance, or add the reset support here. 
___
*[tasks/src/task_model.ts, line 65 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZC8grAAmp2Vkk3XtO:-MZZC8grAAmp2Vkk3XtP:b-3e6yc) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L65)):*
> ```XML
> export interface TaskModelMetadata {
>   /** Model name. */
>   name: string;
> ```

Any suggested naming convention?
___
*[tasks/src/task_model.ts, line 151 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZCKHQ50A91kp7MdYj:-MZZCKHQ50A91kp7MdYk:b-wc971t) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L151)):*
> ```XML
>    * For example, for TFJS mobilenet model, this would be 'mobilenet'.
>    */
>   abstract readonly sourceModelGlobalNs: string;
> ```

sourceModelGlobalNamespace
___
*[tasks/src/task_model.ts, line 172 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZE_0a3DlWMmS0Yr3a:-MZZE_0a3DlWMmS0Yr3b:bp4f7uy) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L172)):*
> ```
> sourceModelGlobal
> ```
Why not call this `sourceModel`?
___
*[tasks/src/task_model.ts, line 213 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZEjPX4P-JtWEt9ZcQ:-MZZEjPX4P-JtWEt9ZcR:bpztcut) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L213)):*
> ```
>    * Loads the source model with the given global namespace and options, and
>    * transforms it to the corresponding task model.
> ```
nit: Loads the source model and transforms it to the corresponding task model.
___
*[tasks/src/tasks/common.ts, line 24 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZFemc4aDtynpS7A3V:-MZZFemc4aDtynpS7A3W:b-mu7t2o) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/common.ts#L24)):*
> ```
> 3.5.0
> ```
3.6.0
___
*[tasks/src/tasks/image_classification/mobilenet_tflite.ts, line 62 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZJJ1J9ZTmH6fnNAv8:-MZZJJ1J9ZTmH6fnNAv9:b-ah9p0o) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/image_classification/mobilenet_tflite.ts#L62)):*
<details><summary><i alt="summary">Quoted 8 lines of code…</i></summary>

> ```
>     if (loadingOptions) {
>       if (loadingOptions.version !== undefined) {
>         mobilenetVersion = String(loadingOptions.version);
>       }
>       if (loadingOptions.alpha !== undefined) {
>         mobilenetAlpha = String(loadingOptions.alpha);
>       }
>     }
> ```
</details>

If you expect every transform needs some option validation, does it make sense to call `this.validateOptions`?

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @mattsoulanille, and @pyu10055)

___
*[tasks/README.md, line 78 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYTeuv3T-SalSbALpC:-MZZKH5z3auyjHVZyrz6:b221iwa) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L78)):*
<details><summary><i>Previously, jinjingforever (Jing Jin) wrote…</i></summary><blockquote>

Are you talking about an example for custom tflite model? If so, there is one above in the "load model and run inference" section.
</blockquote></details>

cool, maybe add a line to say: see example of how TFLite custom models in the 'load model and run inference section'
___
*[tasks/src/package_loader.ts, line 30 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ5UDP-zpfhXWAf-V7:-MZZT4ZQ5jaiiud1o__a:bbyvzw5) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L30)):*

Before the detailed description, it's better to summarize packageUrls allows you to express package dependency, which determines loading order.
</blockquote></details>

If all scripts are loaded sequentially, can this just be a flattened array? I think browser ensure for both web worker and html page, javascripts are evaled in order.
___
*[tasks/src/package_loader.ts, line 41 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZY_DWI2g_ZnX-5HESQ:-MZZSech39YoaFdPg14V:b232byn) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L41)):*
<details><summary><i>Previously, jinjingforever (Jing Jin) wrote…</i></summary><blockquote>

This is the termination condition of the recursive calls. It basically means if there is no more packages to fetch, return the global namespace.
</blockquote></details>

why we need to use recursion, can loop work?
___
*[tasks/src/task_model.ts, line 65 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZC8grAAmp2Vkk3XtO:-MZZS86Q2CBJuMBieeJL:b4494nb) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L65)):*

Any suggested naming convention?
</blockquote></details>

This is model naming, I don't think It is for the class name we discussed. we should not have constraint on it. 
___
*[tasks/src/task_model.ts, line 144 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZLGNbBU0Og5inw9IY:-MZZLGNbBU0Og5inw9IZ:b-4ags1q) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L144)):*
> ```XML
>    * added automatically.
>    */
>   abstract readonly packageUrls: Array<string[]>;
> ```

should this be part of the metadata?
___
*[tasks/src/task_model.ts, line 172 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZLNXaAJGo6ZkOhum0:-MZZLNXaAJGo6ZkOhum1:b-w3gq43) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L172)):*
> ```XML
>   async load(options?: LO): Promise<M> {
>     // Load the packages to get the global namespace ready.
>     const sourceModelGlobal = await this.loadSourceModelGlobalNs(options);
> ```

`loadSourceModelGlobalNamespaces` to be consistent?
___
*[tasks/src/tasks/common.ts, line 76 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYg9UkB6LUQG_ZWdLn:-MZZLj3m8qR7PXulCd4F:b-qa65qk) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/common.ts#L76)):*
<details><summary><i>Previously, jinjingforever (Jing Jin) wrote…</i></summary><blockquote>

> ```
> https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm
> ```

Actually for wasm backend to work, I will have to load tf-backend-wasm script explicitly. If I load "https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm", it will try to load the wasm files from "https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm/" instead of "https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm/dist/" (that is without using setWasmPath).

But you reminded me that I should use minified file here:) Updated.
</blockquote></details>

that is for making `setWasmPath` work? Is this a problem of our wasm package in general or only affecting the dynamic loading logic?
___
*[tasks/src/tasks/image_classification/mobilenet_tfjs.ts, line 79 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYhaej8xY8bw-oOE8k:-MZZQHr42JyVBO0Y7bf8:b1unpns) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/image_classification/mobilenet_tfjs.ts#L79)):*
<details><summary><i>Previously, jinjingforever (Jing Jin) wrote…</i></summary><blockquote>

Looks like it disposes stuff internally. I don't see any public facing dispose method. https://github.com/tensorflow/tfjs-models/blob/master/mobilenet/src/index.ts#L145
</blockquote></details>

I see, we should add one to that package. I will create an issue for that.

jinjingforever (commented):
Thank you for the review Na! 

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-:-MZZjmy_ChI9D0c-pJEK:b4vpsc8)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @mattsoulanille, and @pyu10055)

___
*[tasks/README.md, line 6 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYzGQ632xdVqH6I4_N:-MZZTTfdDTzqGgEQOjYb:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L6)):*

> ```
> web
> ```
Web
</blockquote></details>

Done.
___
*[tasks/README.md, line 9 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYzODI3VkJr6FqrsHw:-MZZUDT305MDtwb4p-1h:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L9)):*

> ```
> TFJS
> ```
Add link to GitHub.
</blockquote></details>

Done.
___
*[tasks/README.md, line 9 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYzSEEA9ru0UB-I2H7:-MZZUDvj3wi1atFp7iCh:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L9)):*

> ```
> TFLite
> ```
Add link to GitHub.
</blockquote></details>

Done.
___
*[tasks/README.md, line 9 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYzTIQ1-vK37WqpUV8:-MZZUEPjDVcKCPKmYwIU:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L9)):*

> ```
> MediaPipe
> ```
Add link to GitHub.
</blockquote></details>

Done.
___
*[tasks/README.md, line 10 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYzZcO2jKyliSMPr1b:-MZZb7LYD9lsn951b4Fi:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L10)):*

> ```
> The model interfaces are specifically designed
> for each task. 
> ```
This sentence is confusing, hard to understand what message it's trying to convey. Consider removing or rephrase.
</blockquote></details>

Done.
___
*[tasks/README.md, line 11 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ-7ae1GaBiq93fouz:-MZZb7iy0lRHacLifwob:b-ars3qg) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L11)):*

> ```
> for each task. They are intuitive and easy-to-use, even for developers without
> any comprehensive ML knowledge. In addition, the API will automatically load
> required packages on the fly. You never need to worry about missing dependencies
> again.
> ```
I understand this part and it is great, but for first time reader, it's hard to follow all those points and the messages are not very clear. Try to write more constructively. What about something like:
The Task API is designed with ease-of-use in mind, it has below features:
1. Models are grouped by ML tasks, making them easy to discover.
2. Simple installation, you only need to install this API without needing to worry about all the models, runtimes, backends.
3. Lightweight package, only 9.8kb. Models, runtimes and backends are only dynamically loaded when needed, and will only be loaded once per page load.
</blockquote></details>

Great point! I combined the comments from you and Ping and rewrote the first paragraph with a shorter overview and 3 bullet points. Thanks!
___
*[tasks/README.md, line 63 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ259F6yVrv8pf7me2:-MZZbwdj1ZVaaUt6IRjP:b-fpjuuw) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L63)):*

> ```
> classify
> ```
Is it possible to simplify the API to `load`, `predict` for all task models? Using `classify` in one task and `estimate` in another task basically have the same effect, which is run prediction, right? TFLite task API uses `classify`, `estimate` etc., so does our tfjs-models. ML5.js uses `predict` for every task. MediaPipe uses `send` for every task. So there are examples in both leagues, but personally I prefer the latter, just a simple API, easy to use, easy to remember, for any users.
</blockquote></details>

I see your points. It does make things easier to use because users don't need to "guess" the method name for different tasks. WDYT @pyu10055 ?
___
*[tasks/README.md, line 86 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ3FNw1raSmgF_sFcr:-MZZdERc-lwxMpnc_Z2Z:b-qd0qdm) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L86)):*

Some of the contents here are pretty advanced, it is contrary to what said above, for users who are not familiar with ML. Maybe separate the contents with a notice or section head, 'More Advanced Contents for Eager Readers'.
</blockquote></details>

Good idea! Done
___
*[tasks/src/package_loader.ts, line 30 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ5UDP-zpfhXWAf-V7:-MZZdjeuDn9CS2p9zQQH:batzxrw) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L30)):*

If all scripts are loaded sequentially, can this just be a flattened array? I think browser ensure for both web worker and html page, javascripts are evaled in order.
</blockquote></details>

Good idea. Done

To Ping: actually the packages in an inner array will be downloaded in parallel. Once they are downloaded, they can be evaluated in any order. This is faster than downloading everything in sequence.
___
*[tasks/src/package_loader.ts, line 36 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ54rR71xjtYWqx6ZG:-MZZehdG4PpAUtKWbYnb:b-qqqx93) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L36)):*

It's better to add somewhere here that this method allows multi-level nested arrays, such as [[[url1, url2], url3], url4]
</blockquote></details>

The type definition of pacakgeUrls is Array<string[]> so it actually doesn't support multi-level nested arrays. Each (tail) recursion call below handles one inner array (a package set).
___
*[tasks/src/package_loader.ts, line 41 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZY_DWI2g_ZnX-5HESQ:-MZZgVNK60UlDMGawBMz:blj57b8) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L41)):*

why we need to use recursion, can loop work?
</blockquote></details>

Good question! I tried the loop, but the combination of Promise.all + await doesn't work as expected. Probably related to how code is transpiled to es5 (didn't look into it).
___
*[tasks/src/package_loader.ts, line 44 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ5vTu8QBOuGLSi8bW:-MZZh-dt8HVRMnUJTC94:b7n66om) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L44)):*

Does it handle namespace conflict? What if namespace already exists?
</blockquote></details>

Good question. For now, we reply on the loaded packages to handle this (like our tfjs-* packages). Also, I think popular bundlers should also be able to handle this automatically. For example, when initializing a "namespace", they should always check if it exists or not. If it exists, re-use it. If not, use {}. (I can see this in rollup's output)
___
*[tasks/src/package_loader.ts, line 66 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ8sfqC5UUY5jEaMUu:-MZZiUUu7k3PgapsW3pW:b-9al67v) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/package_loader.ts#L66)):*

You've researched security issue, right? Any known risk? Do we need to mention it in the README?
</blockquote></details>

Yes. I don't think users should worry about this because **we** will make sure that all loaded packages are CORS friendly (which is true for jsdelivr).

Your comment reminded me that I should add "crossorigin" attribute to the script tags to make them future-proof (for the sharedarraybuffer migration). Thanks!
___
*[tasks/src/package_loader.ts, line 38 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZBoQNCdw9N4D85DyI:-MZZgNU7-9AFqeRJuxBy:bpqrkty) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/package_loader.ts#L38)):*

This doccomment doesn't have parameter description, please add. What is the naming convention for namespace?
</blockquote></details>

Nice catch! Added to the class doc.

jinjingforever (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683)* status: 0 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, and @pyu10055)

___
*[tasks/src/task_model.ts, line 72 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ9sLW6Tmk8fmcDgpQ:-MZZmTeO27_5kMaBomgA:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/task_model.ts#L72)):*

> ```
>  so it makes more sense to put them here
> ```
This is design details, no need to expose it in doccomment.
</blockquote></details>

Done.
___
*[tasks/src/task_model.ts, line 59 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZJqBXCt98yYEbihrv:-MZZl7iL9NegNXGdAs5I:byt1ggf) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L59)):*

Some pipelines have global states (all MP models and pose detection models), and has a method to reset states, how to support that? Either ask user to clean up and recreate an instance, or add the reset support here. 
</blockquote></details>

Good point. Since this is only for certain models, it might make more sense to add this extra method in those models' loader implementations (and we will also need to mention this in the docs). 
___
*[tasks/src/task_model.ts, line 65 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZC8grAAmp2Vkk3XtO:-MZZlwW2B9eFHMNDidLn:b-evc70o) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L65)):*

This is model naming, I don't think It is for the class name we discussed. we should not have constraint on it. 
</blockquote></details>

+1. This is mostly for display purpose in playground or other advanced apps. I am adding "human-readable" to the comment to make it slightly clearer.
___
*[tasks/src/task_model.ts, line 144 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZLGNbBU0Og5inw9IY:-MZZmWbq3Nsk5pORQw5J:b-fuk31n) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L144)):*

should this be part of the metadata?
</blockquote></details>

I thought about it, but I feel like this is the "under-the-hood" kind of data that shouldn't expose to users, so I left it outside.
___
*[tasks/src/task_model.ts, line 151 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZCKHQ50A91kp7MdYj:-MZZn2Lw1X3REiBwtao4:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L151)):*

sourceModelGlobalNamespace
</blockquote></details>

Done.
___
*[tasks/src/task_model.ts, line 172 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZE_0a3DlWMmS0Yr3a:-MZZn3sC4gb9vapJDnRg:brn1fc8) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L172)):*

> ```
> sourceModelGlobal
> ```
Why not call this `sourceModel`?
</blockquote></details>

They are slightly different:) For example, "mobilenet" is the namespace, but the model is actually "mobilenet.Mobilenet". I guess in my mind, I consider "model" to be the "model class", which lives inside the "model namespace" (which might also contain other stuff other than the model class, such as util functions, types, etc). 
___
*[tasks/src/task_model.ts, line 172 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZLNXaAJGo6ZkOhum0:-MZZnvIr69i4NLNR08D_:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f59648ab18f362b3a40669eb4846c456421af30/tasks/src/task_model.ts#L172)):*

`loadSourceModelGlobalNamespaces` to be consistent?
</blockquote></details>

Done.
___
*[tasks/src/tasks/common.ts, line 24 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZFemc4aDtynpS7A3V:-MZZo78UEFRoNluqx2G2:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/common.ts#L24)):*

> ```
> 3.5.0
> ```
3.6.0
</blockquote></details>

Done.
___
*[tasks/src/tasks/common.ts, line 76 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYg9UkB6LUQG_ZWdLn:-MZZo9qu4c0EMjxaI8pT:bnpqd9p) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/common.ts#L76)):*

that is for making `setWasmPath` work? Is this a problem of our wasm package in general or only affecting the dynamic loading logic?
</blockquote></details>

I think this is by design. If you look at the instructions in our [wasm backend page](https://github.com/tensorflow/tfjs/tree/master/tfjs-backend-wasm), it also asks users to import ".../dist/tf-backend-wasm.js" through the script tag. This is the "correct" way to load things without using setWasmPath. 

If you use "https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm" in the script tag, you will need to call `setWasmPath("https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-backend-wasm/dist")` (or something like that) in your script to make things work. The root cause is that the imported js script will look for wasm files from the same location/path of the script.
___
*[tasks/src/tasks/image_classification/mobilenet_tfjs.ts, line 79 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZYhaej8xY8bw-oOE8k:-MZZphdT8JqCvaPEd1ta:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/image_classification/mobilenet_tfjs.ts#L79)):*

I see, we should add one to that package. I will create an issue for that.
</blockquote></details>

Done.
___
*[tasks/src/tasks/image_classification/mobilenet_tflite.ts, line 62 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZJJ1J9ZTmH6fnNAv8:-MZZqZSfBiO0dh_s9V6y:blbt62u) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/src/tasks/image_classification/mobilenet_tflite.ts#L62)):*

> ```
>     if (loadingOptions) {
>       if (loadingOptions.version !== undefined) {
>         mobilenetVersion = String(loadingOptions.version);
>       }
>       if (loadingOptions.alpha !== undefined) {
>         mobilenetAlpha = String(loadingOptions.alpha);
>       }
>     }
> ```
If you expect every transform needs some option validation, does it make sense to call `this.validateOptions`?
</blockquote></details>

I think for most cases, the transformer could just take the model's options type and directly use it without doing anything extra. But I will keep an eye on this part and see if it is necessary to extract it out. Thanks!

pyu10055 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @mattsoulanille, and @pyu10055)

___
*[tasks/README.md, line 63 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ259F6yVrv8pf7me2:-MZnDGn9-5bm94fjz7n5:bgh12hg) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L63)):*
<details><summary><i>Previously, jinjingforever (Jing Jin) wrote…</i></summary><blockquote>

I see your points. It does make things easier to use because users don't need to "guess" the method name for different tasks. WDYT @pyu10055 ?
</blockquote></details>

I agree, we should at least support an conventional method that common across all models. If we want to be align with TFlite task API, we can add sugar on top.
___
*[tasks/src/package_loader.ts, line 41 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZnDpVy2Dn0nG0MC02I:-MZnDpVy2Dn0nG0MC02J:b-h1y9u2) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9c3871fab56eb3fc16ee2999f87e53e8c58c30aa/tasks/src/package_loader.ts#L41)):*
> ```XML
>    * For example, if packageUrls = [ [url1, url2], [url3] ], then url1 and
>    * url2 are loaded together first. url3 will then be loaded when url1 and
>    * url2 are both done loading.
> ```

It will be good to document the different orders of download and js loading by browser. 
i.e. url1 and url2 will be downloaded in parallel, but url1 will loaded in a sequential order.
___
*[tasks/src/package_loader.ts, line 52 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZnEfoSDXVcgF-1GE2q:-MZnEfoSDXVcgF-1GE2r:bsu8r52) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9c3871fab56eb3fc16ee2999f87e53e8c58c30aa/tasks/src/package_loader.ts#L52)):*
> ```XML
>       // tslint:disable-next-line:no-any
>       const global: any = isWebWorker() ? self : window;
>       return global[namespace] as G;
> ```

How is the namespace attached? are we assuming the loaded js file sets up the namespace corerctly?

jinjingforever (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683)* status: 0 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, and @pyu10055)

___
*[tasks/README.md, line 63 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZZ259F6yVrv8pf7me2:-MZnQamY9Kwhm15ccjkb:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/972879b770c3f4e618bf8f3af2f030d902a96ad7/tasks/README.md#L63)):*

I agree, we should at least support an conventional method that common across all models. If we want to be align with TFlite task API, we can add sugar on top.
</blockquote></details>

Done.
___
*[tasks/src/package_loader.ts, line 41 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZnDpVy2Dn0nG0MC02I:-MZnTyDj2fa5E2mrhWnS:btahuc1) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9c3871fab56eb3fc16ee2999f87e53e8c58c30aa/tasks/src/package_loader.ts#L41)):*

It will be good to document the different orders of download and js loading by browser. 
i.e. url1 and url2 will be downloaded in parallel, but url1 will loaded in a sequential order.
</blockquote></details>

Done. I updated the comment in `TaskModelLoader.packageUrls`, added a pointer here.
___
*[tasks/src/package_loader.ts, line 52 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZnEfoSDXVcgF-1GE2q:-MZnUKmr9L4y-coKVwfA:bt2lw9x) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9c3871fab56eb3fc16ee2999f87e53e8c58c30aa/tasks/src/package_loader.ts#L52)):*

How is the namespace attached? are we assuming the loaded js file sets up the namespace corerctly?
</blockquote></details>

Correct. We expect that the dependency packages should get the namespace ready. If not, we will throw an error.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683)* status: 0 of 1 approvals obtained (waiting on @jinjingforever, @lina128, @mattsoulanille, and @pyu10055)

___
*[tasks/README.md, line 85 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZnVfMO-96pzkJYnAaC:-MZnVfMO-96pzkJYnAaD:b-ar7t7y) ([raw file](https://github.com/tensorflow/tfjs-models/blob/75d771d70b6f31c752a13b3a5d9acabf2052f99e/tasks/README.md#L85)):*
> ```
> classify
> ```
model1.predict

jinjingforever (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683)* status: 0 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, and @pyu10055)

___
*[tasks/README.md, line 85 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZnVfMO-96pzkJYnAaC:-MZnWxqN7gx3baty6gaT:byvsbfs) ([raw file](https://github.com/tensorflow/tfjs-models/blob/75d771d70b6f31c752a13b3a5d9acabf2052f99e/tasks/README.md#L85)):*

> ```
> classify
> ```
model1.predict
</blockquote></details>

Good catch! Updated here and the sentences above.

pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @jinjingforever, @lina128, and @mattsoulanille)

___
*[tasks/src/package_loader.ts, line 52 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZnEfoSDXVcgF-1GE2q:-MZnugoZ4eMquTGYMpo5:b-annp1b) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9c3871fab56eb3fc16ee2999f87e53e8c58c30aa/tasks/src/package_loader.ts#L52)):*
<details><summary><i>Previously, jinjingforever (Jing Jin) wrote…</i></summary><blockquote>

Correct. We expect that the dependency packages should get the namespace ready. If not, we will throw an error. 
</blockquote></details>

I don't see how we verify the namespace is set?

jinjingforever (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/683)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128, @mattsoulanille, and @pyu10055)

___
*[tasks/src/package_loader.ts, line 52 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/683#-MZnEfoSDXVcgF-1GE2q:-MZo2i-G1G7U32cg52fv:b-swrt21) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9c3871fab56eb3fc16ee2999f87e53e8c58c30aa/tasks/src/package_loader.ts#L52)):*

I don't see how we verify the namespace is set?
</blockquote></details>

Nice catch!:) I thought I had it in TaskModelLoader but it is not there.. I added the throw here. Thanks!

PR #702, opened 2021-05-10
[pose-detection]Refactor ModelConfig and EstimationConfig.

lina128:
This PR has below changes:
1. Remove quantBytes as a common config for ModelConfig, because this is only used in PoseNet. In the future, if BlazePose and MoveNet want to provide quantized model, the custom modelUrl field should be sufficient.
2. Move enableSmoothing in BlazePose from estimationConfig to modelConfig, because smoothing filter has global state,and it cannot be changed between estimate calls.

### DIFF
--- a/pose-detection/src/blazepose/constants.ts
+++ b/pose-detection/src/blazepose/constants.ts
@@ -35,10 +35,15 @@ export const BLAZEPOSE_DETECTOR_ANCHOR_CONFIGURATION = {
   aspectRatios: [1.0],
   fixedAnchorSize: true
 };
+export const DEFAULT_BLAZEPOSE_MODEL_CONFIG = {
+  lite: false,
+  enableSmoothing: true,
+  detectorModelUrl: DEFAULT_BLAZEPOSE_DETECTOR_MODEL_URL,
+  landmarkModelUrl: DEFAULT_BLAZEPOSE_LANDMARK_MODEL_URL
+};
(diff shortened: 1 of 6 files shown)


pyu10055 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/702)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128 and @mhays-google)

___
*[pose-detection/src/blazepose/constants.ts, line 38 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/702#-M_Mg09LCl57DGoIBCqm:-M_Mg09LCl57DGoIBCqn:b-fb81xe) ([raw file](https://github.com/tensorflow/tfjs-models/blob/6d6ccf67f976658d71a1eaf902393beaa71680db/pose-detection/src/blazepose/constants.ts#L38)):*
> ```XML
>   fixedAnchorSize: true
> };
> export const DEFAULT_BLAZEPOSE_MODEL_CONFIG = {
> ```

should these consts have type declared so it can be type checked.

mhays-google (approved):
> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/702)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @lina128)

PR #703, opened 2021-05-11
[pose-detection]Add model type to BlazePose.

lina128:
This PR adds model type (possible value: lite|full|heavy) to BlazePose. Also disabled static image test because there's a bug that needs to be fixed in @mediapipe/pose.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/703)

### DIFF
--- a/pose-detection/src/blazepose_tfjs/detector.ts
+++ b/pose-detection/src/blazepose_tfjs/detector.ts
@@ -17,6 +17,7 @@
 
 import * as tfconv from '@tensorflow/tfjs-converter';
 import * as tf from '@tensorflow/tfjs-core';
+import {BlazePoseModelType} from '../blazepose_mediapipe/types';
 
 import {SECOND_TO_MICRO_SECONDS} from '../calculators/constants';
 import {convertImageToTensor} from '../calculators/convert_image_to_tensor';
@@ -75,7 +76,8 @@ export class BlazePoseTfjsDetector extends BasePoseDetector {
   private constructor(
       private readonly detectorModel: tfconv.GraphModel,
       private readonly landmarkModel: tfconv.GraphModel,
-      private readonly enableSmoothing: boolean) {
+      private readonly enableSmoothing: boolean,
+      private readonly modelType: BlazePoseModelType) {
     super();
 
     this.anchors =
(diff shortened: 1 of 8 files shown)


pose-detection/src/blazepose_tfjs/detector.ts
```diff
@@ -326,14 +328,42 @@ export class BlazePoseTfjsDetector extends BasePoseDetector {
...
         this.landmarkModel.predict(imageValueShifted) as tf.Tensor[];
 
     let landmarkTensor, poseFlagTensor, heatmapTensor;
 
-    landmarkTensor = landmarkResult[3] as tf.Tensor2D;
-    poseFlagTensor = landmarkResult[0] as tf.Tensor2D;
-    heatmapTensor = landmarkResult[2] as tf.Tensor4D;
+    switch (this.modelType) {
+      case 'lite':
+        landmarkTensor = landmarkResult[1] as tf.Tensor2D;
```
bazarevsky, line 348: In general, it is better to have unified order of output tensors for models, if converter supports this.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/703)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @bazarevsky and @pyu10055)

___
*[pose-detection/src/blazepose_tfjs/constants.ts, line 27 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/703#-M_RKLix9GFTd8I1cpV4:-M_RN65QE3c4m9RWHwr_:bvsol1j) ([raw file](https://github.com/tensorflow/tfjs-models/blob/0b14697b32f28f92dd24428af437b74b665cdd10/pose-detection/src/blazepose_tfjs/constants.ts#L27)):*

when we will have the models on TFHub?
</blockquote></details>

Later.
___
*[pose-detection/src/blazepose_tfjs/detector.ts, line 348 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/703#-M_OnpYTLjbG3c2GOSas-r1-348:-M_RRsOnBbw9IZePeXL3:bm725e) ([raw file](https://github.com/tensorflow/tfjs-models/blob/0b14697b32f28f92dd24428af437b74b665cdd10/pose-detection/src/blazepose_tfjs/detector.ts#L348)):*
<details><summary><i>Previously, bazarevsky (v.bazarevsky) wrote…</i></summary><blockquote>

In general, it is better to have unified order of output tensors for models, if converter supports this. 
</blockquote></details>

Agreed. Currently converter doesn't support it. We are working on supporting named output.

PR #747, opened 2021-06-02
Fix face landmark detection demo

jinjingforever:
This is related to tensorflow/tfjs#5137. This PR includes the following fixes:

- Fixed `ReferenceError: e is not defined` error. It was caused by `parcel` bundler's minification process which is not "compatible" with our WASM backend. The fix is to use rollup with the correct terser config (see [here](https://github.com/tensorflow/tfjs/tree/master/tfjs-backend-wasm#js-minification) for more info).
- The backend switching in the demo was not working correctly. In addition to calling `cancelAnimationFrame`, we also need to stop the currently running callback before setting the new backend (`cancelAnimationFrame` will only cancel the scheduled callback). Without doing this, the "backend not ready" error will be thrown when switching to the WASM backend even if we are using `await tf.setBackend`.
-  Update `face-landmarks-detection` dependency to a specific version instead of a local reference. Using the reference confuses the build tool. I guess the reason is that `face-landmarks-detection` also has a local reference to facemesh. This "double local references" might be problematic.
- Update formatting to demo/index.js. 

Tested locally. Will push a new demo after this PR is merged. Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/747)

### DIFF
--- a/face-landmarks-detection/demo/README.md
+++ b/face-landmarks-detection/demo/README.md
@@ -18,7 +18,13 @@ Install dependencies and prepare the build directory:
 yarn
 ```
 
-To watch files for changes, and launch a dev server:
+Build the model locally which the demo depends on:
+
+```sh
+yarn build-deps
+```
+
+Launch a development server, and watch files for changes.
 
 ```sh
 yarn watch
@@ -26,31 +32,14 @@ yarn watch
 
 ## If you are developing face-landmarks-detection locally, and want to test the changes in the demo
 
-Cd into the face-landmarks-detection folder:
+Cd into the face-landmarks-detection/demo folder:
 ```sh
-cd face-landmarks-detection
+cd face-landmarks-detection/demo
 ```
 
-Install dependencies:
+Rebuild the model locally:
 ```sh
-yarn
-```
-
-Publish face-landmarks-detection locally:
-```sh
-yarn build && yarn yalc publish
-```
-
-Cd into the demo and install dependencies:
-
-```sh
-cd demo
-yarn
-```
-
-Link the local face-landmarks-detection to the demo:
-```sh
-yarn yalc link @tensorflow-models/face-landmarks-detection
+yarn build-deps
 ```
 
 Start the dev demo server:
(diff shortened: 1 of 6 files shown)


pyu10055 (changes requested):
thanks

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/747#-:-MbhbbjO-cDTQsK8WftX:b-k73z76)* status: 0 of 1 approvals obtained (waiting on @jinjingforever)

___
*[face-landmarks-detection/demo/package.json, line 21 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/747#-Mbhbe6-2ztLnlux2uAB:-Mbhbe6-2ztLnlux2uAC:b-nvv9b8) ([raw file](https://github.com/tensorflow/tfjs-models/blob/08b710409092798315404dac1b34fa60cc4fd1a2/face-landmarks-detection/demo/package.json#L21)):*
> ```JSON
>   },
>   "scripts": {
>     "watch": "cross-env NODE_ENV=development parcel index.html --no-hmr --target browser --open",
> ```

looks like `yarn watch` still uses parcel? should it be removed? how will users start the demo? You might need to update the demo README to reflect the changes.

jinjingforever (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/747)* status: 0 of 1 approvals obtained (waiting on @pyu10055)

___
*[face-landmarks-detection/demo/package.json, line 21 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/747#-Mbhbe6-2ztLnlux2uAB:-Mbhlobs7GyT1Rhftv5b:b-14vgue) ([raw file](https://github.com/tensorflow/tfjs-models/blob/08b710409092798315404dac1b34fa60cc4fd1a2/face-landmarks-detection/demo/package.json#L21)):*

looks like `yarn watch` still uses parcel? should it be removed? how will users start the demo? You might need to update the demo README to reflect the changes.
</blockquote></details>

Thanks! Your comment made me realize that changing the `face-landmarks-detection` dependency to a fixed version does not work well with local development. So I changed it back to "link:../dist", and did some more digging. `parcel watch` doesn't work with 'link:../dist', but esbuild does. So I changed `yarn watch` to using esbuild instead. Also, I tweaked the rollup config a little so it now works with "link:../dist" too. 

I also updated the README.md file. 

Thanks!

PR #763, opened 2021-07-26
Add webgpu backend for demos in pose-detection

haoyunfeix:


PixelsCommander:
@haoyunfeix did it ever worked? After multiple tries and tweaks I am still not able to run this

PR #764, opened 2021-07-27
Tracker

ronnyvotel:
Introducing a base class for person tracking.
Most of the simple mechanics are available. The next step is to implement a greedy optimizer, so that detections can be linked with tracks.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764)

### DIFF
--- a/pose-detection/src/calculators/interfaces/common_interfaces.ts
+++ b/pose-detection/src/calculators/interfaces/common_interfaces.ts
@@ -1,4 +1,5 @@
 import {Keypoint} from '../../types';
+import {BoundingBox} from './shape_interfaces';
 
 /**
  * @license
@@ -42,4 +43,13 @@ export interface KeypointsFilter {
   apply(landmarks: Keypoint[], microSeconds: number, objectScale: number):
       Keypoint[];
   reset(): void;
+}
+
+export interface Track {
+  id: number;  // A unique identifier for each tracker person.
+  lastTimestamp: number;  // The last timestamp (in milliseconds) in which a
+                          // detection was linked with the track.
+  keypoints?: Keypoint[];  // The most recent set of keypoints.
+  boundingBox?: BoundingBox;  // The most recent bounding box.
+  score?: number;  // A confidence value of the track.
 }
--- a/pose-detection/src/calculators/interfaces/common_interfaces.ts
+++ b/pose-detection/src/calculators/interfaces/common_interfaces.ts
@@ -46,10 +46,11 @@ export interface KeypointsFilter {
 }
 
 export interface Track {
-  id: number;  // A unique identifier for each tracker person.
+  id: number;  // A unique identifier for each tracked person.
   lastTimestamp: number;  // The last timestamp (in milliseconds) in which a
                           // detection was linked with the track.
-  keypoints?: Keypoint[];  // The most recent set of keypoints.
-  boundingBox?: BoundingBox;  // The most recent bounding box.
+  keypoints?: Keypoint[];  // Keypoints associated with the tracked person.
+  boundingBox?: BoundingBox;  // Bounding box associated with the tracked
+                              // person.
   score?: number;  // A confidence value of the track.
 }
--- a/pose-detection/src/calculators/tracker.ts
+++ b/pose-detection/src/calculators/tracker.ts
@@ -45,6 +45,7 @@ export abstract class Tracker {
   apply(
       poses: Pose[], timestamp: number): Pose[] {
     const simMatrix = this.computeSimilarity(poses);
+    this.filterOldTracks(timestamp);
     this.assignTracks(poses, simMatrix, timestamp);
     this.updateTracks(timestamp);
     return poses;
@@ -59,6 +60,16 @@ export abstract class Tracker {
    */
   abstract computeSimilarity(
       poses: Pose[]): number[][];
+
+  /**
+   * Filters tracks based on their age.
+   * @param timestamp The current timestamp in milliseconds.
+   */
+  filterOldTracks(timestamp: number): void {
+    this.tracks = this.tracks.filter(track => {
+    return timestamp - track.lastTimestamp <= this.maxAge;
+    });
+  }
 
   /**
    * Performs an optimization to link detections with tracks. The `poses`
@@ -77,20 +88,14 @@ export abstract class Tracker {
   /**
    * Updates the stored tracks in the tracker. Specifically, the following
    * operations are applied in order:
-   * 1. Tracks that have not been linked in the past `maxAge` milliseconds are
-   *    removed.
-   * 2. Tracks are sorted based on freshness (i.e. the most recently linked
+   * 1. Tracks are sorted based on freshness (i.e. the most recently linked
    *    tracks are placed at the beginning of the array and the most stale are
    *    at the end).
-   * 3. The tracks array is sliced to only contain `maxTracks` tracks (i.e. the
+   * 2. The tracks array is sliced to only contain `maxTracks` tracks (i.e. the
    *    most fresh tracks).
    * @param timestamp The current timestamp in milliseconds.
    */
   updateTracks(timestamp: number): void {
-    this.tracks = this.tracks.filter(track => {
-      return timestamp - track.lastTimestamp <= this.maxAge;
-    });
-
     // Sort tracks from most recent to most stale, and then only keep the top
     // `maxTracks` tracks.
     this.tracks.sort((ta, tb) => tb.lastTimestamp - ta.lastTimestamp);
--- a/pose-detection/src/calculators/tracker.ts
+++ b/pose-detection/src/calculators/tracker.ts
@@ -25,12 +25,12 @@ import {TrackerConfig} from './interfaces/config_interfaces';
  */
 export abstract class Tracker {
   private tracks: Track[];
-  private readonly max_tracks: number;
-  private readonly max_age: number;
+  private readonly maxTracks: number;
+  private readonly maxAge: number;
 
   constructor(config: TrackerConfig) {
-    this.max_tracks = config.max_tracks;
-    this.max_age = config.max_age;
+    this.maxTracks = config.maxTracks;
+    this.maxAge = config.maxAge;
   }
 
   /**
@@ -43,7 +43,7 @@ export abstract class Tracker {
       poses: Pose[], timestamp: number): Pose[] {
     const sim_matrix = this.computeSimilarity(poses);
     this.assignTracks(poses, sim_matrix);
-    this.removeStaleTracks(timestamp);
+    this.updateTracks(timestamp);
     return poses;
   }
 
@@ -94,18 +94,32 @@ export abstract class Tracker {
 
 
   /**
-   * Removes tracks that have not been linked with poses for a length of time.
+   * Updates the stored tracks in the tracker. Specifically, the following
+   * operations are applied in order:
+   * 1. Tracks that have not been linked in the past `maxAge` milliseconds are
+   *    removed.
+   * 2. Tracks are sorted based on freshness (i.e. the most recently linked
+   *    tracks are placed at the beginning of the array and the most stale are
+   *    at the end).
+   * 3. The tracks array is sliced to only contain `maxTracks` tracks (i.e. the
+   *    most fresh tracks).
+   * @param timestamp The current timestamp in milliseconds.
    */
-  removeStaleTracks(timestamp): void {
+  updateTracks(timestamp: number): void {
     this.tracks = this.tracks.filter(track => {
-      return timestamp - track.lastTimestamp < this.max_age;
+      return timestamp - track.lastTimestamp < this.maxAge;
     })
+
+    // Sort tracks from most recent to most stale, and then only keep the top
+    // `maxTracks` tracks.
+    this.tracks.sort((ta, tb) => tb.lastTimestamp - ta.lastTimestamp);
+    this.tracks = this.tracks.slice(0, this.maxTracks);
   }
 
   /**
    * Removes specific tracks, based on their ids.
    */
-  remove(...ids): void {
+  remove(...ids: number[]): void {
     this.tracks = this.tracks.filter(track => !ids.includes(track.id));
   }
 
--- a/pose-detection/src/calculators/tracker.ts
+++ b/pose-detection/src/calculators/tracker.ts
@@ -44,8 +44,8 @@ export abstract class Tracker {
    */
   apply(
       poses: Pose[], timestamp: number): Pose[] {
-    const simMatrix = this.computeSimilarity(poses);
     this.filterOldTracks(timestamp);
+    const simMatrix = this.computeSimilarity(poses);
     this.assignTracks(poses, simMatrix, timestamp);
     this.updateTracks(timestamp);
     return poses;
--- a/pose-detection/src/calculators/tracker.ts
+++ b/pose-detection/src/calculators/tracker.ts
@@ -51,7 +51,8 @@ export abstract class Tracker {
    * Computes pairwise similarity scores between detections and tracks, based
    * on detected features.
    * @param poses An array of detected `Pose`s.
-   * @returns A 2D array of shape [num_det, num_tracks] with pairwise similarity scores between detections and tracks.
+   * @returns A 2D array of shape [num_det, num_tracks] with pairwise
+   * similarity scores between detections and tracks.
    */
   computeSimilarity(
       poses: Pose[]): number[][] {
--- a/pose-detection/src/calculators/tracker.ts
+++ b/pose-detection/src/calculators/tracker.ts
@@ -41,8 +41,8 @@ export abstract class Tracker {
    */
   apply(
       poses: Pose[], timestamp: number): Pose[] {
-    const sim_matrix = this.computeSimilarity(poses);
-    this.assignTracks(poses, sim_matrix);
+    const simMatrix = this.computeSimilarity(poses);
+    this.assignTracks(poses, simMatrix);
     this.updateTracks(timestamp);
     return poses;
   }
@@ -56,21 +56,21 @@ export abstract class Tracker {
    */
   computeSimilarity(
       poses: Pose[]): number[][] {
-    const num_poses = poses.length;
-    const num_tracks = this.tracks.length;
-    if (!(num_poses && num_tracks)) {
+    const numPoses = poses.length;
+    const numTracks = this.tracks.length;
+    if (!(numPoses && numTracks)) {
       return [[]];
     }
 
-    const sim_matrix = [];
-    for (let i = 0; i < num_poses; ++i) {
-      const sim_row = [];
-      for (let j = 0; j < num_tracks; ++j) {
-        sim_row.push(this.similarityFn(poses[i], this.tracks[j]));
+    const simMatris = [];
+    for (let i = 0; i < numPoses; ++i) {
+      const simRow = [];
+      for (let j = 0; j < numTracks; ++j) {
+        simRow.push(this.similarityFn(poses[i], this.tracks[j]));
       }
-      sim_matrix.push(sim_row);
+      simMatris.push(simRow);
     }
-    return sim_matrix;
+    return simMatris;
   }
 
   /**
@@ -87,12 +87,13 @@ export abstract class Tracker {
    * array is updated in place by providing an `id` property. If incoming 
    * detections are not linked with existing tracks, new tracks will be created.
    * @param poses An array of detected `Pose's.
+   * @param simMatrix A 2D array of shape [num_det, num_tracks] with pairwse
+   * similarity scores between detections and tracks.
    */
   assignTracks(
-      poses: Pose[], similarity_matrix: number[][]): void {
+      poses: Pose[], simMatrix: number[][]): void {
     //TODO: Implement optimization and track store mechanics.
   }
-
 
   /**
    * Updates the stored tracks in the tracker. Specifically, the following
@@ -109,7 +110,7 @@ export abstract class Tracker {
   updateTracks(timestamp: number): void {
     this.tracks = this.tracks.filter(track => {
       return timestamp - track.lastTimestamp < this.maxAge;
-    })
+    });
 
     // Sort tracks from most recent to most stale, and then only keep the top
     // `maxTracks` tracks.
@@ -124,7 +125,7 @@ export abstract class Tracker {
     this.tracks = this.tracks.filter(track => !ids.includes(track.id));
   }
 
-   /**
+  /**
    * Resets tracks.
    */
   reset(): void {
--- a/pose-detection/src/calculators/tracker.ts
+++ b/pose-detection/src/calculators/tracker.ts
@@ -15,7 +15,9 @@
  * =============================================================================
  */
 
-import {Pose, Track} from '../types';
+import {Pose} from '../types';
+import {validateTrackerConfig} from './tracker_utils';
+import {Track} from './interfaces/common_interfaces';
 import {TrackerConfig} from './interfaces/config_interfaces';
 
 /**
@@ -29,6 +31,7 @@ export abstract class Tracker {
   private readonly maxAge: number;
 
   constructor(config: TrackerConfig) {
+    validateTrackerConfig(config);
     this.maxTracks = config.maxTracks;
     this.maxAge = config.maxAge;
   }
@@ -109,7 +112,7 @@ export abstract class Tracker {
    */
   updateTracks(timestamp: number): void {
     this.tracks = this.tracks.filter(track => {
-      return timestamp - track.lastTimestamp < this.maxAge;
+      return timestamp - track.lastTimestamp <= this.maxAge;
     });
 
     // Sort tracks from most recent to most stale, and then only keep the top
--- a/pose-detection/src/calculators/tracker.ts
+++ b/pose-detection/src/calculators/tracker.ts
@@ -45,7 +45,7 @@ export abstract class Tracker {
   apply(
       poses: Pose[], timestamp: number): Pose[] {
     const simMatrix = this.computeSimilarity(poses);
-    this.assignTracks(poses, simMatrix);
+    this.assignTracks(poses, simMatrix, timestamp);
     this.updateTracks(timestamp);
     return poses;
   }
@@ -57,44 +57,20 @@ export abstract class Tracker {
    * @returns A 2D array of shape [num_det, num_tracks] with pairwise
    * similarity scores between detections and tracks.
    */
-  computeSimilarity(
-      poses: Pose[]): number[][] {
-    const numPoses = poses.length;
-    const numTracks = this.tracks.length;
-    if (!(numPoses && numTracks)) {
-      return [[]];
-    }
-
-    const simMatris = [];
-    for (let i = 0; i < numPoses; ++i) {
-      const simRow = [];
-      for (let j = 0; j < numTracks; ++j) {
-        simRow.push(this.similarityFn(poses[i], this.tracks[j]));
-      }
-      simMatris.push(simRow);
-    }
-    return simMatris;
-  }
-
-  /**
-   * Computes the similarity between a single detection and track.
-   * @param pose A single `Pose`.
-   * @returns A scalar which represents the similarity between the given pose
-   * detection and track.
-   */
-  abstract similarityFn(
-      pose: Pose, track: Track): number;
+  abstract computeSimilarity(
+      poses: Pose[]): number[][];
 
   /**
    * Performs an optimization to link detections with tracks. The `poses`
    * array is updated in place by providing an `id` property. If incoming 
    * detections are not linked with existing tracks, new tracks will be created.
    * @param poses An array of detected `Pose's.
-   * @param simMatrix A 2D array of shape [num_det, num_tracks] with pairwse
+   * @param simMatrix A 2D array of shape [num_det, num_tracks] with pairwise
    * similarity scores between detections and tracks.
+   * @param timestamp The current timestamp in milliseconds.
    */
   assignTracks(
-      poses: Pose[], simMatrix: number[][]): void {
+      poses: Pose[], simMatrix: number[][], timestamp: number): void {
     //TODO: Implement optimization and track store mechanics.
   }
 
--- a/pose-detection/src/calculators/tracker.ts
+++ b/pose-detection/src/calculators/tracker.ts
@@ -1,0 +1,118 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+import {Pose, Track} from '../types';
+import {TrackerConfig} from './interfaces/config_interfaces';
+
+/**
+ * A stateful tracker for associating detections between frames. This is an
+ * abstract base class that performs generic mechanics. Implementations must
+ * inherit from this class.
+ */
+export abstract class Tracker {
+  private tracks: Track[];
+  private readonly max_tracks: number;
+  private readonly max_age: number;
+
+  constructor(config: TrackerConfig) {
+    this.max_tracks = config.max_tracks;
+    this.max_age = config.max_age;
+  }
+
+  /**
+   * Tracks person instances across frames based on detections.
+   * @param poses An array of detected `Pose`s.
+   * @param timestamp The timestamp associated with the incoming poses.
+   * @returns An updated array of `Pose`s with tracking id properties.
+   */
+  apply(
+      poses: Pose[], timestamp: number): Pose[] {
+    const sim_matrix = this.computeSimilarity(poses);
+    this.assignTracks(poses, sim_matrix);
+    this.removeStaleTracks(timestamp);
+    return poses;
+  }
+
+  /**
+   * Computes pairwise similarity scores between detections and tracks, based
+   * on detected features.
+   * @param poses An array of detected `Pose`s.
+   * @returns A 2D array of shape [num_det, num_tracks] with pairwise similarity scores between detections and tracks.
+   */
+  computeSimilarity(
+      poses: Pose[]): number[][] {
+    const num_poses = poses.length;
+    const num_tracks = this.tracks.length;
+    if (!(num_poses && num_tracks)) {
+      return [[]];
+    }
+
+    const sim_matrix = [];
+    for (let i = 0; i < num_poses; ++i) {
+      const sim_row = [];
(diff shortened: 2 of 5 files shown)


ronnyvotel:
@googlebot I signed it!

lina128 (commented):
Hi @ronnyvotel, thank you very much for the contribution! Overall it looks great, I left some minor comments below.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-:-Mfdbh9d8l_FCEIWX498:btexlf0)* status: 0 of 1 approvals obtained (waiting on @ronnyvotel)

___
*[pose-detection/src/types.ts, line 79 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfctKfUBu9gxSLg_Sek:-MfctKfUBu9gxSLg_Sel:b-nfjl2e) ([raw file](https://github.com/tensorflow/tfjs-models/blob/424d54825730ef6f962c3fdf14c141a8f386d4de/pose-detection/src/types.ts#L79)):*
<details><summary><i alt="summary">Quoted 7 lines of code…</i></summary>

> ```
> export interface Track {
>   id: number;
>   lastTimestamp: number;
>   keypoints?: Keypoint[];
>   bbox?: BoundingBox;
>   score?: number;
> }
> ```
</details>

The types defined in `src/types.ts` are all exposed to users. I think Tracker is an internal concept, same as Smoothers, for those, they are under `src/calculators`, their type definitions are under `src/calculators/interfaces`.

Please also add annotation for each field.
___
*[pose-detection/src/types.ts, line 81 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfdZ6jB6bUTPUecQsPm:-MfdZ6jB6bUTPUecQsPn:b4z6hgi) ([raw file](https://github.com/tensorflow/tfjs-models/blob/424d54825730ef6f962c3fdf14c141a8f386d4de/pose-detection/src/types.ts#L81)):*
> ```
> bbox
> ```
nit: boundingBox
___
*[pose-detection/src/calculators/tracker.ts, line 32 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfdbIdi0W06hPFaXkVh:-MfdbIdi0W06hPFaXkVi:bwycsu3) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1ca3d3b80c2f4bc082999fbf0780217dd7b019cb/pose-detection/src/calculators/tracker.ts#L32)):*
> ```
> config.maxTracks
> ```
verify maxTracks is non-negative.
___
*[pose-detection/src/calculators/tracker.ts, line 33 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-Mfd_jpqCiI5sffjOLgR:-Mfd_jpsAh7-SaLYmUIH:b-4bslzd) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1ca3d3b80c2f4bc082999fbf0780217dd7b019cb/pose-detection/src/calculators/tracker.ts#L33)):*
> ```
> config.maxAge
> ```
Do you need to verify the configs? e.g. maxAge should not be a negative.
___
*[pose-detection/src/calculators/tracker.ts, line 79 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfdKN0U6uPVSvzwCWTp:-MfdKN0U6uPVSvzwCWTq:b98mym5) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1ca3d3b80c2f4bc082999fbf0780217dd7b019cb/pose-detection/src/calculators/tracker.ts#L79)):*
> ```
> scalar
> ```
If the similarityFn is a model, it's possible that the model may want to run in batches. p.s. Our experience of running a ranking model with TFJS is that running 10 records as a batch is much faster than running 10 model inferences in parallel. Should we just leave the similarityFn part to subclass to decide? Also, the function can be async if it uses a model and needs to download data from GPU to CPU. Change it to async will make it more generic.
___
*[pose-detection/src/calculators/tracker.ts, line 112 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfdYN3j9zXvfURtVUDP:-MfdYN3k7_WspLjI6Wjp:b-g1mvsd) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1ca3d3b80c2f4bc082999fbf0780217dd7b019cb/pose-detection/src/calculators/tracker.ts#L112)):*
> ```
> <
> ```
`<=`?

ronnyvotel (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764)* status: 0 of 1 approvals obtained (waiting on @ardoerlemans and @lina128)

___
*[pose-detection/src/calculators/tracker.ts, line 32 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfdbIdi0W06hPFaXkVh:-MfeOqr_4tdBS9-L-hBq:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1ca3d3b80c2f4bc082999fbf0780217dd7b019cb/pose-detection/src/calculators/tracker.ts#L32)):*

> ```
> config.maxTracks
> ```
verify maxTracks is non-negative.
</blockquote></details>

Done.
___
*[pose-detection/src/calculators/tracker.ts, line 33 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-Mfd_jpqCiI5sffjOLgR:-MfeOrN9D9VwlP8xKPTC:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1ca3d3b80c2f4bc082999fbf0780217dd7b019cb/pose-detection/src/calculators/tracker.ts#L33)):*

> ```
> config.maxAge
> ```
Do you need to verify the configs? e.g. maxAge should not be a negative.
</blockquote></details>

Done.
___
*[pose-detection/src/calculators/tracker.ts, line 79 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfdKN0U6uPVSvzwCWTp:-MfdrQVD-6HtEZV1UDYB:bra0by6) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1ca3d3b80c2f4bc082999fbf0780217dd7b019cb/pose-detection/src/calculators/tracker.ts#L79)):*

> ```
> scalar
> ```
If the similarityFn is a model, it's possible that the model may want to run in batches. p.s. Our experience of running a ranking model with TFJS is that running 10 records as a batch is much faster than running 10 model inferences in parallel. Should we just leave the similarityFn part to subclass to decide? Also, the function can be async if it uses a model and needs to download data from GPU to CPU. Change it to async will make it more generic.
</blockquote></details>

Yeah I think we leave it up to the subclasses. I believe we shouldn't force async here (I'm not too sure whether typescript will even allow "async abstract"). WDYT?
___
*[pose-detection/src/calculators/tracker.ts, line 112 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfdYN3j9zXvfURtVUDP:-MfeHUpK3lFmpgklvQnO:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1ca3d3b80c2f4bc082999fbf0780217dd7b019cb/pose-detection/src/calculators/tracker.ts#L112)):*

> ```
> <
> ```
`<=`?
</blockquote></details>

Done.
___
*[pose-detection/src/types.ts, line 79 at r2](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfctKfUBu9gxSLg_Sek:-Mfdqiuh0yjR1XQ7XAZJ:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/424d54825730ef6f962c3fdf14c141a8f386d4de/pose-detection/src/types.ts#L79)):*

> ```
> export interface Track {
>   id: number;
>   lastTimestamp: number;
>   keypoints?: Keypoint[];
>   bbox?: BoundingBox;
>   score?: number;
> }
> ```
The types defined in `src/types.ts` are all exposed to users. I think Tracker is an internal concept, same as Smoothers, for those, they are under `src/calculators`, their type definitions are under `src/calculators/interfaces`.

Please also add annotation for each field.
</blockquote></details>

Done.

pose-detection/src/calculators/interfaces/common_interfaces.ts
```diff
@@ -43,3 +44,12 @@ export interface KeypointsFilter {
       Keypoint[];
   reset(): void;
 }
+
+export interface Track {
+  id: number;  // A unique identifier for each tracker person.
```
ardoerlemans, line 49: tracked person?

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764)* status: 0 of 1 approvals obtained (waiting on @ardoerlemans, @lina128, and @ronnyvotel)

___
*[pose-detection/src/calculators/tracker.ts, line 79 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfdKN0U6uPVSvzwCWTp:-MfiUzbW3PI_9oilz8co:bsbmuju) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1ca3d3b80c2f4bc082999fbf0780217dd7b019cb/pose-detection/src/calculators/tracker.ts#L79)):*
<details><summary><i>Previously, ronnyvotel wrote…</i></summary><blockquote>

Yeah I think we leave it up to the subclasses. I believe we shouldn't force async here (I'm not too sure whether typescript will even allow "async abstract"). WDYT?
</blockquote></details>

What about batch mode? If the model is going to run in batch mode, it will expect a pose and all trackers. Should we just leave computeSimlarity abstract?

pose-detection/src/calculators/interfaces/common_interfaces.ts
```diff
@@ -43,3 +44,12 @@ export interface KeypointsFilter {
...
+  lastTimestamp: number;  // The last timestamp (in milliseconds) in which a
+                          // detection was linked with the track.
+  keypoints?: Keypoint[];  // The most recent set of keypoints.
```
ardoerlemans, line 52: Do you want to force trackers to use only the most recent set of keypoints and bounding box? I guess a slightly more advanced tracker would want to keep some history to make predictions of where it expects keypoints and bounding boxes.

pose-detection/src/calculators/tracker_utils.ts
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+ import {TrackerConfig} from './interfaces/config_interfaces';
+
+ export function validateTrackerConfig(config: TrackerConfig): void {
+  if (config.maxTracks < 0) {
```
ardoerlemans, line 21: I guess setting maxTracks to 0 also isn't very useful?

pose-detection/src/calculators/tracker_utils.ts
```diff
@@ -0,0 +1,31 @@
...
+    throw new Error(
+      `Must specify 'maxTracks' to be non-negative, but ` +
+      `encountered ${config.maxTracks}`);
+  }
+  if (config.maxAge < 0) {
```
ardoerlemans, line 26: If we allow maxAge == 0, then we should probably update the code that filters based on age, because it will cause all tracks to be filtered out for every frame.

pose-detection/src/calculators/tracker.ts
```diff
@@ -0,0 +1,137 @@
...
+   */
+  abstract similarityFn(
+      pose: Pose, track: Track): number;
+
+  /**
+   * Performs an optimization to link detections with tracks. The `poses`
+   * array is updated in place by providing an `id` property. If incoming 
+   * detections are not linked with existing tracks, new tracks will be created.
+   * @param poses An array of detected `Pose's.
+   * @param simMatrix A 2D array of shape [num_det, num_tracks] with pairwse
```
ardoerlemans, line 93: pairwise

pose-detection/src/calculators/tracker.ts
```diff
@@ -0,0 +1,137 @@
...
+   */
+  computeSimilarity(
+      poses: Pose[]): number[][] {
+    const numPoses = poses.length;
+    const numTracks = this.tracks.length;
+    if (!(numPoses && numTracks)) {
+      return [[]];
+    }
+
+    const simMatris = [];
```
ardoerlemans, line 68: simMatrix

pose-detection/src/calculators/tracker.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  /**
+   * Performs an optimization to link detections with tracks. The `poses`
+   * array is updated in place by providing an `id` property. If incoming 
+   * detections are not linked with existing tracks, new tracks will be created.
+   * @param poses An array of detected `Pose's.
+   * @param simMatrix A 2D array of shape [num_det, num_tracks] with pairwse
+   * similarity scores between detections and tracks.
+   */
+  assignTracks(
```
ardoerlemans, line 96: This also needs a timestamp parameter to update the lastTimestamp in a track.

ardoerlemans (commented):
Looks great! 

I have a few comments.

ronnyvotel (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764)* status: 0 of 1 approvals obtained (waiting on @ardoerlemans and @lina128)

___
*[pose-detection/src/calculators/tracker.ts, line 79 at r3](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfdKN0U6uPVSvzwCWTp:-Mfj8rl28IaAOT0XFmmI:bpau30j) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1ca3d3b80c2f4bc082999fbf0780217dd7b019cb/pose-detection/src/calculators/tracker.ts#L79)):*

What about batch mode? If the model is going to run in batch mode, it will expect a pose and all trackers. Should we just leave computeSimlarity abstract?
</blockquote></details>

Yep, I see your point. I am making computeSimilarity() abstract.
___
*[pose-detection/src/calculators/tracker.ts, line 68 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfcnxBI5Qjud939XADk-r4-68:-Mfj8ehg38H0Dw6WOrld:boj1qgt) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker.ts#L68)):*
<details><summary><i>Previously, ardoerlemans (Ard Oerlemans) wrote…</i></summary><blockquote>

simMatrix
</blockquote></details>

Good catch. Done.
___
*[pose-detection/src/calculators/tracker.ts, line 93 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfcnxBI5Qjud939XADk-r4-93:-Mfj91HR-Thikh-JjPwl:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker.ts#L93)):*
<details><summary><i>Previously, ardoerlemans (Ard Oerlemans) wrote…</i></summary><blockquote>

pairwise
</blockquote></details>

Done.
___
*[pose-detection/src/calculators/tracker.ts, line 96 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfcnxBI5Qjud939XADk-r4-96:-Mfj9L8_C5R7WdobUCGh:b-jpkv8k) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker.ts#L96)):*
<details><summary><i>Previously, ardoerlemans (Ard Oerlemans) wrote…</i></summary><blockquote>

This also needs a timestamp parameter to update the lastTimestamp in a track.
</blockquote></details>

Good call. Added.
___
*[pose-detection/src/calculators/tracker_utils.ts, line 21 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfeMkbBXepJUZ5rrb71-r4-21:-Mfj9Q2qDJjRNMZ7LDDm:bbaomxn) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker_utils.ts#L21)):*
<details><summary><i>Previously, ardoerlemans (Ard Oerlemans) wrote…</i></summary><blockquote>

I guess setting maxTracks to 0 also isn't very useful?
</blockquote></details>

Alright, going to enforce that this is strictly positive.
___
*[pose-detection/src/calculators/tracker_utils.ts, line 26 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfeMkbBXepJUZ5rrb71-r4-26:-MfjANka-L_GW0TuB58s:bjjg027) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker_utils.ts#L26)):*
<details><summary><i>Previously, ardoerlemans (Ard Oerlemans) wrote…</i></summary><blockquote>

If we allow maxAge == 0, then we should probably update the code that filters based on age, because it will cause all tracks to be filtered out for every frame.
</blockquote></details>

OK, going to make sure maxAge is strictly positive
___
*[pose-detection/src/calculators/interfaces/common_interfaces.ts, line 49 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfeMkbBXepJUZ5rrb70-r4-49:-MfjAbwxAs4YVKZR5dMk:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/interfaces/common_interfaces.ts#L49)):*
<details><summary><i>Previously, ardoerlemans (Ard Oerlemans) wrote…</i></summary><blockquote>

tracked person?
</blockquote></details>

Done.
___
*[pose-detection/src/calculators/interfaces/common_interfaces.ts, line 52 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfeMkbBXepJUZ5rrb70-r4-52:-MfjAgC_6wKrnlrxTnix:bxj7yw0) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/interfaces/common_interfaces.ts#L52)):*
<details><summary><i>Previously, ardoerlemans (Ard Oerlemans) wrote…</i></summary><blockquote>

Do you want to force trackers to use only the most recent set of keypoints and bounding box? I guess a slightly more advanced tracker would want to keep some history to make predictions of where it expects keypoints and bounding boxes.
</blockquote></details>

Good point, I updated the comment. In general we will not enforce how these fields are populated.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764)* status: 0 of 1 approvals obtained (waiting on @ardoerlemans and @lina128)

___
*[pose-detection/src/calculators/tracker.ts, line 96 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfcnxBI5Qjud939XADk-r4-96:-MfjCy5y0kn6phMmjzLb:b-73m6vw) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker.ts#L96)):*
<details><summary><i>Previously, ronnyvotel wrote…</i></summary><blockquote>

Good call. Added.
</blockquote></details>

Nice catch!
___
*[pose-detection/src/calculators/tracker_utils.ts, line 21 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfeMkbBXepJUZ5rrb71-r4-21:-MfjDqP0Ao5Nw6T5VadO:bewbj5c) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker_utils.ts#L21)):*
<details><summary><i>Previously, ronnyvotel wrote…</i></summary><blockquote>

Alright, going to enforce that this is strictly positive.
</blockquote></details>

What happens when maxTracks is smaller than maxPoses? Do we allow it?
___
*[pose-detection/src/calculators/tracker_utils.ts, line 26 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfeMkbBXepJUZ5rrb71-r4-26:-MfjDIIwAzsaIe-BFzsx:bizzffj) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker_utils.ts#L26)):*
<details><summary><i>Previously, ronnyvotel wrote…</i></summary><blockquote>

OK, going to make sure maxAge is strictly positive
</blockquote></details>

I thought <= 0 won't filter out all the tracks, because the tracks that just got its lastTimeStamp field updated with the timestamp will have 0 when comparing timestamp - tracker.lastTimeStamp, which is truthy and will be retained in the tracker list.

lina128 (approved):
Thank you for the contribution Ronny! LGTM! I left two comments/questions in the tracker_utils.ts file, just for my own understanding.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-:-MfjEFeP46cxdzlzxNqw:bwqqv54)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ardoerlemans and @lina128)

ardoerlemans (approved):
Thanks for the updates!

ronnyvotel (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @ardoerlemans and @lina128)

___
*[pose-detection/src/calculators/tracker_utils.ts, line 21 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfeMkbBXepJUZ5rrb71-r4-21:-MfjH5s6AtS-cmoARe50:b-88iqsm) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker_utils.ts#L21)):*

What happens when maxTracks is smaller than maxPoses? Do we allow it?
</blockquote></details>

Yeah I think we allow it, since someone will have to actively change defaults. I **believe** that if someone supplies maxPoses larger than maxTracks, we will still return up to maxPoses detections, but the tracked ids will at most stay consistent on only maxTracks people. I'd rather keep the two configurations decoupled, and expect that a user is only updating configs when they have read the documentation and know what they are doing.
___
*[pose-detection/src/calculators/tracker_utils.ts, line 26 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfeMkbBXepJUZ5rrb71-r4-26:-MfjIxzvEubtjrXGzw64:b-fju1s4) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker_utils.ts#L26)):*

I thought <= 0 won't filter out all the tracks, because the tracks that just got its lastTimeStamp field updated with the timestamp will have 0 when comparing timestamp - tracker.lastTimeStamp, which is truthy and will be retained in the tracker list.
</blockquote></details>

That's a great point. Now that I think about it, perhaps we want to do the following on each step:
- filter out tracks based on age (current timestamp relative to lastTimestamp)
- do the assignment (i.e. matching between detections and tracks)
- sort and clip the number of tracks to maxTracks

That way we truly enforce that we never use tracks that are past their age limit, and we always maintain at most maxTracks tracks. WDYT?

ronnyvotel (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @ardoerlemans and @lina128)

___
*[pose-detection/src/calculators/tracker.ts, line 48 at r6](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfjS0Nd1CMdsrXvBPq2:-MfjS0Nd1CMdsrXvBPq3:bxrzaz5) ([raw file](https://github.com/tensorflow/tfjs-models/blob/d26c27f671dc3726dfe644b7b0785c149e7b3d6f/pose-detection/src/calculators/tracker.ts#L48)):*
> ```
> this.filterOldTracks(timestamp);
> ```
Take a look at this new flow and let me know if it works.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @ardoerlemans and @lina128)

___
*[pose-detection/src/calculators/tracker_utils.ts, line 21 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfeMkbBXepJUZ5rrb71-r4-21:-MfjRpDQ2H97W58MYcoS:b-jtqtgj) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker_utils.ts#L21)):*
<details><summary><i>Previously, ronnyvotel wrote…</i></summary><blockquote>

Yeah I think we allow it, since someone will have to actively change defaults. I **believe** that if someone supplies maxPoses larger than maxTracks, we will still return up to maxPoses detections, but the tracked ids will at most stay consistent on only maxTracks people. I'd rather keep the two configurations decoupled, and expect that a user is only updating configs when they have read the documentation and know what they are doing.
</blockquote></details>

Should we add this explanation to the maxTracks annotation? e.g. Consider set maxTracks to be equal or larger than maxPoses. And also details what happens if maxTracks is smaller than the number of returned poses. 
___
*[pose-detection/src/calculators/tracker_utils.ts, line 26 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfeMkbBXepJUZ5rrb71-r4-26:-MfjSlT23xPFHF_jRqA5:bue3qbf) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker_utils.ts#L26)):*
<details><summary><i>Previously, ronnyvotel wrote…</i></summary><blockquote>

That's a great point. Now that I think about it, perhaps we want to do the following on each step:
- filter out tracks based on age (current timestamp relative to lastTimestamp)
- do the assignment (i.e. matching between detections and tracks)
- sort and clip the number of tracks to maxTracks

That way we truly enforce that we never use tracks that are past their age limit, and we always maintain at most maxTracks tracks. WDYT?
</blockquote></details>

This can't handle pause well, although it will reduce comparison. It's a tradeoff. I don't have a strong preference, which aspect do you think is more important? Support paused video or reduce comparison?

ardoerlemans (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @ardoerlemans, @lina128, and @ronnyvotel)

___
*[pose-detection/src/calculators/tracker.ts, line 48 at r6](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfjS0Nd1CMdsrXvBPq2:-MfjU7_W4SgkywmA4aHW:b-qh1r9z) ([raw file](https://github.com/tensorflow/tfjs-models/blob/d26c27f671dc3726dfe644b7b0785c149e7b3d6f/pose-detection/src/calculators/tracker.ts#L48)):*
<details><summary><i>Previously, ronnyvotel wrote…</i></summary><blockquote>

> ```
> this.filterOldTracks(timestamp);
> ```
Take a look at this new flow and let me know if it works.
</blockquote></details>

Maybe filter the tracks before you calculate the similarity matrix? Otherwise you may end up with a matrix that references tracks that have already been filtered out.

ronnyvotel (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @ardoerlemans, @lina128, and @ronnyvotel)

___
*[pose-detection/src/calculators/tracker_utils.ts, line 21 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfeMkbBXepJUZ5rrb71-r4-21:-MfjW8kR-pWReJ8_l4yU:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker_utils.ts#L21)):*

Should we add this explanation to the maxTracks annotation? e.g. Consider set maxTracks to be equal or larger than maxPoses. And also details what happens if maxTracks is smaller than the number of returned poses. 
</blockquote></details>

Done.
___
*[pose-detection/src/calculators/tracker_utils.ts, line 26 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfeMkbBXepJUZ5rrb71-r4-26:-MfjWE8iB6HksM--aFkT:bsu4q23) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker_utils.ts#L26)):*

This can't handle pause well, although it will reduce comparison. It's a tradeoff. I don't have a strong preference, which aspect do you think is more important? Support paused video or reduce comparison?
</blockquote></details>

I think we want to make sure that we aren't ever linking with a track that is stale. In the current implementation we might actually violate the maxAge, and then still link it and assign the current time. I don't like that behavior. I think the paused video case is something that an end developer must work around themselves (by providing the correct timestamps)

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @ardoerlemans, @lina128, and @ronnyvotel)

___
*[pose-detection/src/calculators/tracker.ts, line 83 at r7](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfjYAaBBGmYu_hT6d6u:-MfjYAaBBGmYu_hT6d6v:b-6c0dnc) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f461c93e2ef256327e3e81c9ecce9e5f0c82e79/pose-detection/src/calculators/tracker.ts#L83)):*
> ```
> assignTracks
> ```
Do you want to pass in a scoreThreshold too?
___
*[pose-detection/src/calculators/tracker_utils.ts, line 26 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfeMkbBXepJUZ5rrb71-r4-26:-MfjX1Nt-JYzDjyWvj94:b32w49) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1f096fa2afd1febeb916ecda5defc15e1e2cedd0/pose-detection/src/calculators/tracker_utils.ts#L26)):*
<details><summary><i>Previously, ronnyvotel wrote…</i></summary><blockquote>

I think we want to make sure that we aren't ever linking with a track that is stale. In the current implementation we might actually violate the maxAge, and then still link it and assign the current time. I don't like that behavior. I think the paused video case is something that an end developer must work around themselves (by providing the correct timestamps)
</blockquote></details>

SG
___
*[pose-detection/src/calculators/interfaces/common_interfaces.ts, line 55 at r7](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfjYSXU9j3cdDNTTeYH:-MfjYSXU9j3cdDNTTeYI:b-7i3j52) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f461c93e2ef256327e3e81c9ecce9e5f0c82e79/pose-detection/src/calculators/interfaces/common_interfaces.ts#L55)):*
> ```
> A confidence value of the track
> ```
Where does it come from? Is it the similarity score? Or is it the Pose score? How is it going to be used?
___
*[pose-detection/src/calculators/interfaces/config_interfaces.ts, line 72 at r7](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfjXq10AVFuJeX8gOMm:-MfjXq10AVFuJeX8gOMn:b-5p5l8s) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f461c93e2ef256327e3e81c9ecce9e5f0c82e79/pose-detection/src/calculators/interfaces/config_interfaces.ts#L72)):*
> ```
> TrackerConfig
> ```
Should confidence threshold be tunable too?

ronnyvotel (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/764)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @ardoerlemans and @lina128)

___
*[pose-detection/src/calculators/tracker.ts, line 48 at r6](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfjS0Nd1CMdsrXvBPq2:-MfkHWco-PuGU7J7ZaHA:buf54yl) ([raw file](https://github.com/tensorflow/tfjs-models/blob/d26c27f671dc3726dfe644b7b0785c149e7b3d6f/pose-detection/src/calculators/tracker.ts#L48)):*
<details><summary><i>Previously, ardoerlemans (Ard Oerlemans) wrote…</i></summary><blockquote>

Maybe filter the tracks before you calculate the similarity matrix? Otherwise you may end up with a matrix that references tracks that have already been filtered out.
</blockquote></details>

good catch. Done.
___
*[pose-detection/src/calculators/tracker.ts, line 83 at r7](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfjYAaBBGmYu_hT6d6u:-MfkHl544qQbXrqTMfiw:bg8c15j) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f461c93e2ef256327e3e81c9ecce9e5f0c82e79/pose-detection/src/calculators/tracker.ts#L83)):*

> ```
> assignTracks
> ```
Do you want to pass in a scoreThreshold too?
</blockquote></details>

For now we can assume that confidence thresholding is being done by the detector, and the whole purpose of the tracker is to create fresh tracks, link tracks with new detections, and kill tracks based on age.
___
*[pose-detection/src/calculators/interfaces/common_interfaces.ts, line 55 at r7](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfjYSXU9j3cdDNTTeYH:-MfkI7EH6xsdmiLXRBK9:b-f7g3eh) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f461c93e2ef256327e3e81c9ecce9e5f0c82e79/pose-detection/src/calculators/interfaces/common_interfaces.ts#L55)):*

> ```
> A confidence value of the track
> ```
Where does it come from? Is it the similarity score? Or is it the Pose score? How is it going to be used?
</blockquote></details>

This is an optional placeholder that isn't currently used, but I like having it exposed in the interface for downstream trackers to use. One potential use case is as follows:
When a new detection is linked with a track, set the score to the detection score. Then start decaying the score in time. Use the score as an additional signal when computing similarities (i.e. link detections not only with spatial information like keypoints/boxes, but also consider the score). WDYT?
___
*[pose-detection/src/calculators/interfaces/config_interfaces.ts, line 72 at r7](https://reviewable.io/reviews/tensorflow/tfjs-models/764#-MfjXq10AVFuJeX8gOMm:-MfkHHD87kjZdGXe73OT:b-b7ohqr) ([raw file](https://github.com/tensorflow/tfjs-models/blob/4f461c93e2ef256327e3e81c9ecce9e5f0c82e79/pose-detection/src/calculators/interfaces/config_interfaces.ts#L72)):*

> ```
> TrackerConfig
> ```
Should confidence threshold be tunable too?
</blockquote></details>

I'm thinking that the score might be used internally by the tracker for assignment, and perhaps won't have hard cutoffs. Purely rejecting detections based on confidence might only be performed in the detector.

PR #766, opened 2021-07-29
[pose-detection]Fix demo.

lina128:
Fix https://github.com/tensorflow/tfjs/issues/5400

The current code depend on specific version of @mediapipe/pose, but the wasm file comes from the latest. It broke the demo today. The temporary fix in this PR is to pin the wasm file to the 0.3.x version. Long term solution is for @mediapipe/pose to export a version variable, so that we can always get a matched wasm file.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/766)

### DIFF
--- a/pose-detection/demos/live_video/src/index.js
+++ b/pose-detection/demos/live_video/src/index.js
@@ -53,7 +53,7 @@ async function createDetector() {
         return posedetection.createDetector(STATE.model, {
           runtime,
           modelType: STATE.modelConfig.type,
-          solutionPath: 'https://cdn.jsdelivr.net/npm/@mediapipe/pose'
+          solutionPath: 'https://cdn.jsdelivr.net/npm/@mediapipe/pose@0.3.1621277220'
         });
       } else if (runtime === 'tfjs') {
         return posedetection.createDetector(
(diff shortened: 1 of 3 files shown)


mhays-google (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/766)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

___
*[pose-detection/demos/live_video/src/index.js, line 56 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/766#-MfoANXo0dfEetrzK6XG:-MfoANXo0dfEetrzK6XH:bp0zvrf) ([raw file](https://github.com/tensorflow/tfjs-models/blob/2a1e35e87628e980066ceab874ac6a7318099b76/pose-detection/demos/live_video/src/index.js#L56)):*
> ```JavaScript
>           runtime,
>           modelType: STATE.modelConfig.type,
>           solutionPath: 'https://cdn.jsdelivr.net/npm/@mediapipe/pose@0.3.1621277220'
> ```

This is okay, but you could also use 0.3.

PR #828, opened 2021-08-30
Add MediaPipe hands implementation

ahmedsabie:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/828)


lina128 (commented):
Given that the APIs will have significant difference from the old one, I suggest we create a new package for all hand-detection models. Also, please break the PR into smaller pieces. The initial checkin should only have high level class definitions and API definitions without implementation details. Once all the reviewers have reviewed the APIs and agreed. Then you can create subsequent PRs, each implement one piece, for example, a subclass, a calculator, etc. Break it down to as small as possible. Otherwise, it will be very hard to review. You can close this PR and create a new one.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/828#-:-MiS7lY620RGFThEF5ia:b-iiwp3m)* status: 0 of 1 approvals obtained

PR #830, opened 2021-08-31
Move pose-detection code needed in hand detection to a common folder

ahmedsabie:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/830)


lina128 (commented):
For the common calculators, have you tried just a shared folder at top level, without creating a new NPM package?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/830#-:-MiT56g_-NZUYKqZPBxd:b-jnprru)* status: 0 of 1 approvals obtained

ahmedsabie:
> For the common calculators, have you tried just a shared folder at top level, without creating a new NPM package?
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/830#-:-MiT56g_-NZUYKqZPBxd:b-jnprru)_ status: 0 of 1 approvals obtained

I was able to get it working when compiling but then Karma wasn't recognizing any files outside the package as a typescript file (only allowed me javascript). I tried to look in the tfjs code as well for examples of common folders being shared but the only sharing happened through packages not a folder

lina128 (commented):
You probably need to change what files to include, something like this: https://github.com/tensorflow/tfjs-models/blob/master/pose-detection/karma.conf.js#L59

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/830#-:-MiTSExb9P5T31mJFwx4:b31zrqp)* status: 0 of 1 approvals obtained

ahmedsabie:
> You probably need to change what files to include, something like this: https://github.com/tensorflow/tfjs-models/blob/master/pose-detection/karma.conf.js#L59
> 
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/830#-:-MiTSExb9P5T31mJFwx4:b31zrqp)_ status: 0 of 1 approvals obtained

Ah thanks for bringing it up, I tried it again and it works now! First time around I tried modifying the files in the conf and I got the only js files are allowed error I mentioned above, but it turns out that I had to also modify the preprocessor line which allows karma to accept typescript

ahmedsabie:
I ended up looking at 4 approaches to have the shared code folder usable by multiple packages:

1) Import directly from the folder and simply add it to the karma.conj.js so testing works, which is what I originally used in the [first commit](https://github.com/tensorflow/tfjs-models/commit/980fce1081cdb2184d22f4c358ffee3a2490c750). This works but the problem is that the resulting npm package will have shared at the top level, and the specific package pushed further down in its own directoy (so importing directly using 'package-name' doesn't work you have to do something like 'package-name/src/package-name').

2) Use references in tsconfig. This had the same result as 1)

3) Using a script to copy files manually, this fixes the npm package structure but makes editing confusing (VSCode would point to the locally copied files but you have to modify the file directly in the root shared directory)

4) Using a symlink in src to the top level shared folder. This both fixes the npm package structure as well as makes editing either the top level shared folder or the local symlinked shared folder valid (as they are both modifying the same file). Only issue is searching in vscode would result in two sets of shared files but that is a simple fix in settings.json of VSCode.

All other approaches I found online required creating new packages or adding more dependencies. Overall approach 4 seemed the easiest with no disadvantages for our current use cases and is what I updated my PR with. Ping brought up that when we end up adding test files we want to make sure it's not re-tested in all the packages the shared folder is included in so that's the one thing to keep in mind when we do add tests

lina128 (commented):
Ahmed, thank you for the exploration, this solution looks great! I only left one question. Also please squash the 2 commits.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/830#-:-Mj3Id90627HMAaWa-QG:bx0nekw)* status: 0 of 1 approvals obtained (waiting on @ahmedsabie and @pyu10055)

___
*[pose-detection/src/test_util.ts, line 125 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/830#-Mj3Gfkn7yNnhDhOun9k:-Mj3Gfkn7yNnhDhOun9l:b-sspw8z) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9a43968cfb2470410c9a2db1b6ddd16efb1a592f/pose-detection/src/test_util.ts#L125)):*
> ```
> poseDetection.Keypoint
> ```
Why this change?

ahmedsabie (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/830)* status: 0 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

___
*[pose-detection/src/test_util.ts, line 125 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/830#-Mj3Gfkn7yNnhDhOun9k:-Mj5NtxEB-ejHISTVwhs:bnpy8uj) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9a43968cfb2470410c9a2db1b6ddd16efb1a592f/pose-detection/src/test_util.ts#L125)):*

> ```
> poseDetection.Keypoint
> ```
Why this change?
</blockquote></details>

The Keypoint is moved to the shared folder so I imported it from there, to keep it under poseDetection import I would have to export it there in index should I do that instead?

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/830)* status: 0 of 1 approvals obtained (waiting on @ahmedsabie and @pyu10055)

___
*[pose-detection/src/test_util.ts, line 125 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/830#-Mj3Gfkn7yNnhDhOun9k:-Mj5fg8p5laNH1m4rw-k:b35fp40) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9a43968cfb2470410c9a2db1b6ddd16efb1a592f/pose-detection/src/test_util.ts#L125)):*
<details><summary><i>Previously, ahmedsabie wrote…</i></summary><blockquote>

The Keypoint is moved to the shared folder so I imported it from there, to keep it under poseDetection import I would have to export it there in index should I do that instead?
</blockquote></details>

Is `Keypoint` not exposed by poseDetection API anymore? The test here is actually testing whether the types are properly exposed.

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/830)* status: 0 of 1 approvals obtained (waiting on @ahmedsabie and @pyu10055)

___
*[pose-detection/src/types.ts, line 20 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/830#-Mj5gMjO9cCHbFzLZYzO:-Mj5gMjO9cCHbFzLZYzP:byqe2tk) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1ce148771ef2f60ff29dde07af9cb1100986c696/pose-detection/src/types.ts#L20)):*
> ```
> import {Keypoint}
> ```
Should we export it? How will pose-detection users who use Typescript to use this type?

ahmedsabie (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/830)* status: 0 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

___
*[pose-detection/src/test_util.ts, line 125 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/830#-Mj3Gfkn7yNnhDhOun9k:-Mj5wBFT5omhhIgfsPCT:b-3pq9zd) ([raw file](https://github.com/tensorflow/tfjs-models/blob/9a43968cfb2470410c9a2db1b6ddd16efb1a592f/pose-detection/src/test_util.ts#L125)):*

Is `Keypoint` not exposed by poseDetection API anymore? The test here is actually testing whether the types are properly exposed.
</blockquote></details>

That's a great point, should work now as before.
___
*[pose-detection/src/types.ts, line 20 at r4](https://reviewable.io/reviews/tensorflow/tfjs-models/830#-Mj5gMjO9cCHbFzLZYzO:-Mj5wHBK3bExqOBWo6PO:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-models/blob/1ce148771ef2f60ff29dde07af9cb1100986c696/pose-detection/src/types.ts#L20)):*

> ```
> import {Keypoint}
> ```
Should we export it? How will pose-detection users who use Typescript to use this type?
</blockquote></details>

Done.

lina128 (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/830#-:-Mj6UkNWDS-z6TZs0iP9:bb6ywmx)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128 and @pyu10055)

lina128:
cc @ardoerlemans

PR #834, opened 2021-09-20
Add mediapipe API for hands

ahmedsabie:
This API will be used in the detector creation function as well as the estimateHands methods of the MediaPipe detector implementation

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/834)

### DIFF
--- a/hand-detection/src/mediapipe/types.ts
+++ b/hand-detection/src/mediapipe/types.ts
@@ -1,0 +1,51 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+import {EstimationConfig, ModelConfig} from '../types';
+
+/**
+ * Common MediaPipeHands model config.
+ */
+export interface MediaPipeHandsModelConfig extends ModelConfig {
+  runtime: 'mediapipe'|'tfjs';
+}
+
+export interface MediaPipeHandsEstimationConfig extends EstimationConfig {}
+
+/**
+ * Model parameters for MediaPipeHands MediaPipe runtime
+ *
+ * `runtime`: Must set to be 'mediapipe'.
+ *
+ * `solutionPath`: Optional. The path to where the wasm binary and model files
+ * are located.
+ *
+ * `minDetectionConfidence`: Optional. Default to 0.5. Minimum confidence value
+ * from the hand detection model for detection to be considered succesful.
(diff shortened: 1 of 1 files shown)


lina128 (changes requested):
The description mentioned detector creation function, but the existing codebase doesn't have this function. Can you add the detector creation function in this PR? Otherwise, the type definition itself is not used anywhere. Please also amend the commits to this commits.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/834#-:-Mk4Bj1V2BBNyTn3cpZY:bdjq7rf)* status: 0 of 1 approvals obtained (waiting on @ahmedsabie and @ivan-grishchenko)

___
*[hand-detection/src/mediapipe/types.ts, line 44 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/834#-Mk4BYSVEZkanexaZT-i:-Mk4BYSVEZkanexaZT-j:b7lrei6) ([raw file](https://github.com/tensorflow/tfjs-models/blob/c97b8fc5067fac879bb4c31e1f5bd94357c46ef4/hand-detection/src/mediapipe/types.ts#L44)):*
> ```
> minDetectionConfidence
> ```
@ivan-grishchenko Should we expose this to user?

ahmedsabie:
> The description mentioned detector creation function, but the existing codebase doesn't have this function. Can you add the detector creation function in this PR? Otherwise, the type definition itself is not used anywhere. Please also amend the commits to this commits.

I don't think the detector creation function should be added in this PR because it requires implementation details of the Mediapipe detector class constructor which should come in the implementation PR. I mentioned the use cases in the PR description as a reference for whats coming in the next PRs but not necessarily what already exists, I can edit it to remove that if its less confusing

lina128 (approved):
Ah, right. SG.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/834#-:-Mk8MBUa4sHjn0dIV3p6:b-sjry5w)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ahmedsabie and @ivan-grishchenko)

lina128 (commented):
LGTM. Please also wait for Ivan's review.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/834#-:-Mk8Mu6W-wk7r4XWHalM:bks6qfk)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ahmedsabie and @ivan-grishchenko)

hand-detection/src/mediapipe/types.ts
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * `runtime`: Must set to be 'mediapipe'.
+ *
+ * `solutionPath`: Optional. The path to where the wasm binary and model files
+ * are located.
+ *
+ * `minDetectionConfidence`: Optional. Default to 0.5. Minimum confidence value
+ * from the hand detection model for detection to be considered succesful.
+ */
+export interface MediaPipeHandsMediaPipeModelConfig extends
```
ivan-grishchenko, line 40: nit: Check how we dealt with duplication of `MediaPipe` in the name in FaceMesh pipeline and do the same here.

ivan-grishchenko (approved):

hand-detection/src/mediapipe/types.ts
ahmedsabie: FaceMesh doesn't support both TFJS and MediaPipe implementation I believe, but it does use the name mediapipeFacemesh which would be the same format we are using here.

PR #836, opened 2021-09-21
Remove minDetectionConfidence from hands config

ahmedsabie:
Removing the minDetectionConfidence due to @lina128 pointing out that users shouldn't need it to reject hands, they can use keypoint score to do so, which is more accurate.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/836)


lina128 (approved):
LGTM

PR #854, opened 2021-10-30
Rename hand-detection package to hand-pose-detection

ahmedsabie:
Name update due to hand-detection only conveying a meaning of detecting bounding boxes rather than poses

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/854)


lina128 (approved):
LGTM

PR #858, opened 2021-11-03
Change sync method to async for blazeface and USE demo

haoyunfeix:


jenperson:
Thank you for your contribution! The Blazeface package has since been deprecated. Can you push a new version that just includes the changes to the Universal Sentence Encoder demo?

PR #867, opened 2021-11-12
Add MediaPipe body segmentation API

ahmedsabie:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/867)


lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/867)* status: 0 of 1 approvals obtained (waiting on @ahmedsabie)

___
*[hand-pose-detection/src/mediapipe/detector.ts, line 18 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/867#-MoL_jqa6E2Gh73ONI-L:-MoL_jqa6E2Gh73ONI-M:bb1xowv) ([raw file](https://github.com/tensorflow/tfjs-models/blob/a55569921401324669b3e879d79cd4fb4653b3ab/hand-pose-detection/src/mediapipe/detector.ts#L18)):*
> ```
> ```
This is a breaking change, please also change mediapipe folder's README's installation section. Also change package.json's version to 1.0.0
___
*[pose-detection/src/blazepose_mediapipe/detector.ts, line 18 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/867#-MoLa5ffEgWbY_GBPcIo:-MoLa5ffEgWbY_GBPcIp:b6b4f4u) ([raw file](https://github.com/tensorflow/tfjs-models/blob/a55569921401324669b3e879d79cd4fb4653b3ab/pose-detection/src/blazepose_mediapipe/detector.ts#L18)):*
> ```
> ```
Same as above, this is a breaking change. Maybe do this change in a separate PR? Either way is fine with me.

ahmedsabie (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/867)* status: 0 of 1 approvals obtained (waiting on @ahmedsabie and @lina128)

___
*[hand-pose-detection/src/mediapipe/detector.ts, line 18 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/867#-MoL_jqa6E2Gh73ONI-L:-MoWz-mI2EgY-JQD3uuF:b-baw9lf) ([raw file](https://github.com/tensorflow/tfjs-models/blob/a55569921401324669b3e879d79cd4fb4653b3ab/hand-pose-detection/src/mediapipe/detector.ts#L18)):*

> ```
> ```
This is a breaking change, please also change mediapipe folder's README's installation section. Also change package.json's version to 1.0.0
</blockquote></details>

Sounds good, also not sure if this matters but right now I am checking for if input is a tensor on line 138 using instanceof, but I could also check the "not" case instead which is checking for existence of width property which would mean that non tensor inputs would not use any tfjs code, would that change dependecies?

lina128 (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/867)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ahmedsabie)

___
*[hand-pose-detection/src/mediapipe/detector.ts, line 18 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/867#-MoL_jqa6E2Gh73ONI-L:-Mou45Ge33e14fVFQFm_:bkgpvoq) ([raw file](https://github.com/tensorflow/tfjs-models/blob/a55569921401324669b3e879d79cd4fb4653b3ab/hand-pose-detection/src/mediapipe/detector.ts#L18)):*
<details><summary><i>Previously, ahmedsabie wrote…</i></summary><blockquote>

Sounds good, also not sure if this matters but right now I am checking for if input is a tensor on line 138 using instanceof, but I could also check the "not" case instead which is checking for existence of width property which would mean that non tensor inputs would not use any tfjs code, would that change dependecies?
</blockquote></details>

Doesn't matter.

PR #898, opened 2022-01-13
Fix timing with GPU

ahmedsabie:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/898)


lina128 (commented):
Thank you Ahmed, the code change for benchmarking looks good! No need to submit this change, because this is only for benchmarking offline. For the actual demo, we don't want to break GPU pipeline as much as possible.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/898#-:-MtPwEsUEoJHG5x5ij4h:b-8ido49)* status: 0 of 1 approvals obtained

PR #911, opened 2022-01-19
[pose-detection] support passing IOHandler to model urls of blazepose_tfjs and movenet

jinjingforever:
Background: teamsportz.pro wants to load movenet model from their react-native app bundle instead of internet. tfjs-react-native already has [BundleResourceIO](https://github.com/tensorflow/tfjs/blob/master/tfjs-react-native/src/bundle_resource_io.ts#L24) which is a `io.IOHandler`, and our converter's `loadGraphModel` already supports it. 

This PR updates the "modelUrl" type to include io.IOHandler, and modified the logic of checking whether the url is from tfhub or not by adding a check to make sure the url is a string. I did this for both blazepose_tfjs and movenet.

Looks like blazepose_tfjs/detector.ts file was not properly formatted. My editor formatted it and it shows lots of diffs here. Please only focus on the `load` function at the end.

Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/911)


lina128 (approved):
Thank you Jing! LGTM. cc @ahmedsabie

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/911#-:-Mto2zmd2VUSCA2fdGIm:b-enq78c)* status: :shipit: complete! 1 of 1 approvals obtained

PR #950, opened 2022-02-15
Fix mediapipe obfuscated package export name

ahmedsabie:
Addresses issue mentioned in https://discuss.tensorflow.org/t/typeerror-when-attempting-to-create-mediapipehands-detector/7602. 

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/950)


mattsoulanille (approved):
> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/950)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @lina128)

___
*[body-segmentation/rollup.config.js, line 55 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/950#-Mvz4H4gAgMt47_nrvVk:-Mvz4H4gAgMt47_nrvVl:b-zdg92q) ([raw file](https://github.com/tensorflow/tfjs-models/blob/53e7f8b7fe667720b9881c3c164f1aa4ff2b86b4/body-segmentation/rollup.config.js#L55)):*
> ```JavaScript
>         '@tensorflow/tfjs-converter': 'tf',
>         // Package is obfuscated so class is directly attached to window.
>         '@mediapipe/selfie_segmentation': 'window'
> ```

This is an odd solution, but I don't know of a better way to do this. Ideally, we would change the `this` that gets passed to selfie_segmentation when it loads to avoid the case where`this == window` and it attaches everything to `window`, but I don't know of a way to do that. 

It looks like they're using the same approach in their (codepen)[https://codepen.io/mediapipe/pen/wvJyQpq?editors=0010]. If you click Settings -> HTML -> Stuff for head, they have a similar script tag as we have:
```
  <script src="https://cdn.jsdelivr.net/npm/@mediapipe/selfie_segmentation@0.1/selfie_segmentation.js" crossorigin="anonymous"></script>
```
So I think this is a reasonable approach given that's how they're loading it in their example.

Also, I was afraid this might break loading from npm, but after thinking about it, I don't think it does. The files should be exactly the same since you're loading a file from the npm package in either case (just from jsdelivr in the http:// case).



_Code quote:_
```JavaScript
        '@mediapipe/selfie_segmentation': 'window'
```

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/950)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[body-segmentation/rollup.config.js, line 55 at r1](https://reviewable.io/reviews/tensorflow/tfjs-models/950#-Mvz4H4gAgMt47_nrvVk:-Mw7vcXY4R6Aqa7bItFe:bsnsco) ([raw file](https://github.com/tensorflow/tfjs-models/blob/53e7f8b7fe667720b9881c3c164f1aa4ff2b86b4/body-segmentation/rollup.config.js#L55)):*
<details><summary><i>Previously, mattsoulanille (Matthew Soulanille) wrote…</i></summary><blockquote>

This is an odd solution, but I don't know of a better way to do this. Ideally, we would change the `this` that gets passed to selfie_segmentation when it loads to avoid the case where`this == window` and it attaches everything to `window`, but I don't know of a way to do that. 

It looks like they're using the same approach in their (codepen)[https://codepen.io/mediapipe/pen/wvJyQpq?editors=0010]. If you click Settings -> HTML -> Stuff for head, they have a similar script tag as we have:
```
  <script src="https://cdn.jsdelivr.net/npm/@mediapipe/selfie_segmentation@0.1/selfie_segmentation.js" crossorigin="anonymous"></script>
```
So I think this is a reasonable approach given that's how they're loading it in their example.

Also, I was afraid this might break loading from npm, but after thinking about it, I don't think it does. The files should be exactly the same since you're loading a file from the npm package in either case (just from jsdelivr in the http:// case).
</blockquote></details>

Please sync with Michael and Tyler. They should expose a global namespace.

lina128 (approved):
Thank you Ahmed!

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/950#-:-Mx6rfBP8fA2HCDsOaH8:bfpkus8)* status: :shipit: complete! 2 of 1 approvals obtained

supermoos:
I'm seeing an error when bundling for production with Vite (vitejs.dev) where the bundled build complains like below. Is this perhaps related to this change?

vendor.4a05ada5.js:35370 Uncaught (in promise) TypeError: pose.Pose is not a constructor
    at new n (vendor.4a05ada5.js:35370:91)
    at vendor.4a05ada5.js:35436:26
    at vendor.4a05ada5.js:35256:20
    at Object.next (vendor.4a05ada5.js:35265:8)
    at vendor.4a05ada5.js:35202:33
    at new Promise (<anonymous>)
    at L (vendor.4a05ada5.js:35181:10)
    at et (vendor.4a05ada5.js:35426:10)
    at vendor.4a05ada5.js:36848:26
    at vendor.4a05ada5.js:35256:20

anselanza:
Just to confirm, this is what I am getting, too. Also trying to use Vite to bundle a project that uses MediaPipe BlazePose and TypeScript. Everything works fine in development, and `vite build` completes without errors, but I get a similar runtime message as below.

> I'm seeing an error when bundling for production with Vite (vitejs.dev) where the bundled build complains like below. Is this perhaps related to this change?
> 
> vendor.4a05ada5.js:35370 Uncaught (in promise) TypeError: pose.Pose is not a constructor at new n

emkooz:
Building with Vite seems to be fixed as of Vite 3.0.0. 

The production build needs to fully use esbuild: checkout ["Using esbuild deps optimization at build time"](https://vitejs.dev/guide/migration.html#experimental).

PR #1038, opened 2022-08-12
Fix blazeface memory leak

sklum:
Hi tfjs team 👋!

This PR fixes two memory leaks in blazeface - hopefully this is helpful even though it's deprecated.

I can confirm (at least in terms of allocated tensors) that there aren't any more leaks.  I also removed an unnecessary null check from the previous PR that added the `dispose()` method.

Is there any chance these changes could be pushed to `npm`?  I'd be happy to switch to the new face detection engine but that detector doesn't return a probability / confidence AFAIK and this functionality is required for my work.

Thanks for all your work and your willingness to open source this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/1038)

### DIFF
--- a/blazeface/src/blazeface_test.ts
+++ b/blazeface/src/blazeface_test.ts
@@ -24,6 +24,8 @@ import * as blazeface from './index';
 import {stubbedImageVals} from './test_util';
 
 describeWithFlags('BlazeFace', NODE_ENVS, () => {
+  const startTensors = tf.memory().numTensors;
+
   let model: BlazeFaceModel;
   beforeAll(async () => {
     model = await blazeface.load();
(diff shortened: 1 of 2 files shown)


ahmedsabie:
@sklum Thanks for the fix! Can you modify [this test](https://github.com/tensorflow/tfjs-models/blob/master/blazeface/src/blazeface_test.ts#L32) to check the startTensors as well? The current test doesn't check for cleanup by the model, only cleanup by inference. Here's an [example](https://github.com/tensorflow/tfjs-models/blob/master/face-detection/src/tfjs/tfjs_test.ts#L51) doing that.

sklum:
Will do!

sklum:
Alright @ahmedsabie, added the test.  If you run the new test without calling `dispose()` within, it will indeed fail.  Calling `dispose()` works as expected.

pyu10055 (approved):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/1038)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @ahmedsabie)

ahmedsabie (approved):

PR #1039, opened 2022-08-16
Allow locateFile to be specified directly

satoren:
What is better about this one is that it allows you to manage files directly in Bundler.

When combined with vite3, it looks like this

```typescript
import {
  BodySegmenter,
  SupportedModels,
  createSegmenter 
} from '@tensorflow-models/body-segmentation'


const selfieSegmentationmodulesMap = async ()=> {
  const modules = import.meta.glob('../../node_modules/@mediapipe/selfie_segmentation/*', { as: 'url' })
  const ret:[string, string][] = []
  for (const entry of Object.entries(modules)) {
    ret.push([entry[0].split('/').pop() || '', await entry[1]()])
  }
  return Object.fromEntries(ret)
}


export const createMediapipeSegmenter = async (): Promise<BodySegmenter> => {
  const modulesMap = await selfieSegmentationmodulesMap()
  const model = SupportedModels.MediaPipeSelfieSegmentation
  return createSegmenter(model, {
    runtime: 'mediapipe',
    modelType: 'general',
    locateFile: (file) => {
      return modulesMap[file]
    }
  })
}

```

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/1039)

### DIFF
--- a/body-segmentation/src/selfie_segmentation_mediapipe/types.ts
+++ b/body-segmentation/src/selfie_segmentation_mediapipe/types.ts
@@ -42,12 +42,15 @@ export interface MediaPipeSelfieSegmentationSegmentationConfig extends
  *
  * `solutionPath`: Optional. The path to where the wasm binary and model files
  * are located.
+ * `locateFile`: Optional. The function to return URLs of the wasm binary and
+ * model files. If specified at the same time as solutionPath,
+ * solutionPath is ignored.
  */
 export interface MediaPipeSelfieSegmentationMediaPipeModelConfig extends
     MediaPipeSelfieSegmentationModelConfig {
   runtime: 'mediapipe';
   solutionPath?: string;
-  locateFile?: (path: string, base: string) => string;
+  locateFile?: (path: string, prefix?: string) => string;
 }
 
 /**
--- a/body-segmentation/src/selfie_segmentation_mediapipe/types.ts
+++ b/body-segmentation/src/selfie_segmentation_mediapipe/types.ts
@@ -47,6 +47,7 @@ export interface MediaPipeSelfieSegmentationMediaPipeModelConfig extends
     MediaPipeSelfieSegmentationModelConfig {
   runtime: 'mediapipe';
   solutionPath?: string;
+  locateFile?: (path: string, base: string) => string;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


body-segmentation/src/selfie_segmentation_mediapipe/types.ts
```diff
@@ -47,6 +47,7 @@ export interface MediaPipeSelfieSegmentationMediaPipeModelConfig extends
     MediaPipeSelfieSegmentationModelConfig {
   runtime: 'mediapipe';
   solutionPath?: string;
+  locateFile?: (path: string, base: string) => string;
```
ahmedsabie, line 50: Please add a comment to the interface describing this option, and similarly in README.md in the same folder as this file.

ahmedsabie (changes requested):

ahmedsabie:
Also there is a lint error: `body-segmentation/src/selfie_segmentation_mediapipe/segmenter.ts[74, 6]: Missing semicolon`

ahmedsabie (commented):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/1039)* status: 0 of 1 approvals obtained (waiting on @satoren)

ahmedsabie (approved):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/1039)* status: :shipit: complete! 1 of 1 approvals obtained

lina128 (commented):
Hi thank you for your contribution @satoren , just want to understand your use case better. Why the existing solutionPath argument can't fulfill your use case?

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/1039#-:-NBsT_M4-PG8CNWSvJz9:biw4spr)* status: :shipit: complete! 1 of 1 approvals obtained

satoren:
Hi thank you for the review,  I'm using [vite](https://vitejs.dev/) for Bundler, but when I import a script with vite, the actual file name contains a hash, so I cannot use it in the solutionPath. Therefore, I had to copy the file to the asset directory manually or automatically (e.g., with a plugin). Also, if the file is copied to the asset directory in that way, the file name does not change between versions, so the old version may be referenced when the cache is set.

lina128 (approved):
Got it, fair enough. LGTM

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-models/1039#-:-NC1lwzpEVEA_eTCVbcX:buit5fe)* status: :shipit: complete! 2 of 1 approvals obtained

PR #1171, opened 2023-07-27
Upgrade body segmentation model

fengwuyao:
Upgrade body segmentation model. Upgrade TFJS version to 4.9.0.

### DIFF
--- a/body-segmentation/package.json
+++ b/body-segmentation/package.json
@@ -47,7 +47,7 @@
     "ts-node": "~8.8.2",
     "tslint": "~6.1.3",
     "tslint-no-circular-imports": "~0.7.0",
-    "typescript": "~4.4.0",
+    "typescript": "~5.0.4",
     "yalc": "~1.0.0-pre.21"
   },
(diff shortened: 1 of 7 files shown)


Linchenn (approved):

body-segmentation/package.json
```diff
@@ -46,7 +47,7 @@
     "ts-node": "~8.8.2",
     "tslint": "~6.1.3",
     "tslint-no-circular-imports": "~0.7.0",
-    "typescript": "~3.9.9",
+    "typescript": "~4.4.0",
```
mattsoulanille, line 50: Nit: We should update to 5.0.4 if possible, but this isn't strictly necessary.

mattsoulanille (approved):

body-segmentation/package.json
mattsoulanille: Actually, that can wait for tfjs 4.10.0 (or can be done now. it shouldn't really affect the demo).

PR #1270, opened 2023-11-09
Testing against tfjs_4.13_0 release

laxmareddyp:


mattsoulanille (approved):
LGTM. We should merge this and the other test PRs when they pass (although we should change the commit name / PR name to something like `Update TFJS version for blazeface, body-pix, and body-segmentation`).

pyu10055 (approved):

PR #1278, opened 2023-11-09
Update TFJS version for coco-ssd, depth estimation, face-detetction and gpt2.

laxmareddyp:


pyu10055:
@laxmareddyp thank you for updating the models, have you confirmed the new versions works for the demos of these models?

laxmareddyp:
> @laxmareddyp thank you for updating the models, have you confirmed the new versions works for the demos of these models?

Will check tomorrow for all the successful build models and update you.Thanks

mattsoulanille (approved):

PR #1320, opened 2023-12-22
universal-sentence-encoder packages update

erikyo:
To use this model with the current version of tfjs it is needed to update this package to a version 4 or higher. for the same reason i updated other dependencies from the tests everything works correctly

Summarizing:
- Package version updated to 1.3.4
- TensorFlow Updated to >4
- Typescript to >5
- some additional dependencies like jasmine, rimraf, ts-node have been updated in order to ensure better functionality and safety

### DIFF
--- a/universal-sentence-encoder/package.json
+++ b/universal-sentence-encoder/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow-models/universal-sentence-encoder",
-  "version": "1.3.3",
+  "version": "1.3.4",
   "description": "Universal Sentence Encoder lite in TensorFlow.js",
   "main": "dist/index.js",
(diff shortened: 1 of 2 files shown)


ram-you:
We really need this update. Is there a collaborator to check this?
@pyu10055

erikyo:
@ram-you completely understand and since I was in the same situation and needed it for [an experiment](https://erikyo.github.io/searchai/) ([example](https://erikyo.github.io/searchai/)) I redid the [USE package](https://www.npmjs.com/package/universal-sentence-encoder-alt) as I needed it. I keep it as long as it is needed and then deprecate it in favor of this one when it is updated (it doesn't matter with this pr or not, I just made it to push it a bit since it really needs an update)

PR #1368, opened 2024-04-02
Update classes.ts

IsaacGeorg:
We have updated certain classes for object detection in beach cleaning mechanism to our project


IsaacGeorg:
I have modified and deleted some of the classes that I want to use datasets in our project called SoBot which is a beach waste cleaning mechanism in which the esp32 camera module helps to identify the type of object (waste) that we need to collect, would you please consider this request to make changes in the dataset for our project.

